PR #364, opened 2015-01-07
Report unhandled rejected promises

panuhorsmalahti:
Report unhandled rejected promises as per rsvp instructions.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -32,7 +32,6 @@ var link = require('./link');
 var asp = require('rsvp').denodeify;
 var System = require('systemjs');
 
-
 var core = module.exports;
 
 exports.run = function(moduleName) {
@@ -63,7 +62,7 @@ exports.build = function() {
   .then(function(doTranspile) {
     if (doTranspile)
       config.pjson.buildConfig.transpileES6 = true;
-    
+
     if (!config.pjson.buildConfig || config.pjson.buildConfig.minify === undefined) {
       saveConfig = true;
       return ui.confirm('Minify?', true);
@@ -111,7 +110,7 @@ exports.setMode = function(modes) {
   .then(function() {
     if (modes.indexOf('local') == -1)
       return true;
-    
+
     // set local
     Object.keys(config.loader.endpoints).forEach(function(e) {
       config.loader.endpoints[e].setLocal();
--- a/lib/install.js
+++ b/lib/install.js
@@ -7,7 +7,6 @@ var PackageName = require('./config/package-name');
 var ui = require('./ui');
 var path = require('path');
 var link = require('./link');
-
 var rimraf = require('rimraf');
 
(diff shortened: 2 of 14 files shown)


guybedford:
Thanks! I think we only need to do this once? Perhaps only in `cli.js` and `api.js`, the two entry points.

panuhorsmalahti:
Yeah, technically it should be only required with the first require, but perhaps it's cleaner to do it always to be certain?

guybedford:
Because it's an event we only need to do it once, otherwise we will have multiple errors.

panuhorsmalahti:
Ah, that's a good point. Moved those to api.js and cli.js. I also cleaned up extra whitespace at the end of lines (it's my editor setting).

lib/core.js
```diff
@@ -14,7 +14,8 @@
  *   limitations under the License.
  */
 
-var Promise = require('rsvp').Promise;
+var RSVP = require('rsvp');
+var Promise = RSVP.Promise;
```
guybedford: This isn't necessary.

lib/install.js
```diff
@@ -1,5 +1,6 @@
 var config = require('./config');
-var Promise = require('rsvp').Promise;
+var RSVP = require('rsvp');
+var Promise = RSVP.Promise;
```
guybedford: Again, not necessary.

guybedford:
Sure thanks, it would be nicer without the whitespace diffs though.

Also, the Promise requires don't need to be changed anymore.

panuhorsmalahti:
Changed the requires.

guybedford:
Great thanks, can you squash the commits?

panuhorsmalahti:
Squashed commits.

guybedford:
I was thinking it may make more sense for this to be a soft failure instead
of crashing everything:

```javascript
ui.log('warn', 'Unhandled rejection \n' + err.stack || err);
```

panuhorsmalahti:
It's better to close the process.

guybedford:
I'm not sure? I don't want to crash an application when it doesn't need to. There is one valid case that unhandled rejections might happen, which is if a fork resolution results in an existing download no longer being relevant. In such a scenario, the download promise is no longer listened to, and may no longer lead to a rejection handler, but that wouldn't be critical.

panuhorsmalahti:
If you don't want the application to crash you should handle the rejection.

guybedford:
@panuhorsmalahti this would involve adding listeners during the point of fork resolution to ensure that unattached download promises have silent rejections.

guybedford:
@panuhorsmalahti thanks so much for driving this, I've separately merged a first implementation in https://github.com/jspm/jspm-cli/commit/fe6af8799d42c827d1f79f6888b86c89cb3e6439.

Issue #336, opened 2014-12-23
jspm register registry function


guybedford:
`jspm register jquery github:components/jquery`

Sends a PR to the jspm registry with a new / modified entry.

trusktr:
This will indeed be great. bower has it's own thing like this, so definitely this will help make things easier. Nonetheless, I do believe there will be cases where people will want to override the registry overrides.

+1 for jspm.overrides in package.json.

trusktr:
Just wanted to note: bower allows devs to add a package to the bower registry without needing a PR (or maybe PRs are automatically merged). It's first come, first serve.

trusktr:
Would `jspm register override` and `jspm register name` replace an item if it already exists?

guybedford:
Perhaps we can consider having additions being sent through an API that ensure they are always merged instantly so we get the same behaviour as Bower and npm.

But a PR would always be needed for overrides or adjustments of existing entries.

trusktr:
Yeah, that'd be nice. Bower has a thing where a package can only be modified by the owner of the git repo, and to do so requires providing an API key.

PR #370, opened 2015-01-08
Support for jspm.overrides in package.json.

trusktr:
Here's a starting point. The command line override takes precedence. Now need to modify pkg.write() so it doesn't get removed from package.json.:wq

### DIFF
--- a/lib/config/package.js
+++ b/lib/config/package.js
@@ -213,6 +213,12 @@ PackageJSON.prototype.write = function() {
       if (this.originalPjson[p] === pjson[p])
         delete pjson[p];
     }
+
+    // keep original overrides
+    try {
+      pjson.overrides = this.originalPjson.jspm.overrides;
+    } catch(e) {}
+
     this.originalPjson.jspm = pjson;
   }
 
--- a/lib/config/package.js
+++ b/lib/config/package.js
@@ -215,9 +215,8 @@ PackageJSON.prototype.write = function() {
     }
 
     // keep original overrides
-    try {
+    if (this.originalPjson.jspm && this.originalPjson.jspm.overrides)
       pjson.overrides = this.originalPjson.jspm.overrides;
-    } catch(e) {}
 
     this.originalPjson.jspm = pjson;
   }
--- a/lib/config/package.js
+++ b/lib/config/package.js
@@ -93,6 +93,7 @@ PackageJSON.prototype.read = function(prompts) {
     self.name = pjson.name || defaults.name;
 
     self.registry = pjson.registry;
+    self.overrides = pjson.overrides;
     self.dependencies = {};
     // only read dependencies when combined with a registry property
     if (pjson.dependencies && pjson.registry) {
@@ -166,6 +167,10 @@ PackageJSON.prototype.write = function() {
   if (this.packages != defaults.packages)
     directories.packages = this.packages;
 
+  console.log('---------------------------------------------')
+  console.log('this', this)
+  console.log('---------------------------------------------')
+
   if (hasProperties(directories)) {
     for (var d in directories)
       directories[d] = winPath(path.relative(this.dir, directories[d]));
@@ -214,9 +219,8 @@ PackageJSON.prototype.write = function() {
         delete pjson[p];
     }
 
-    // keep original overrides
-    if (this.originalPjson.jspm && this.originalPjson.jspm.overrides)
-      pjson.overrides = this.originalPjson.jspm.overrides;
+    // save the original overrides
+    pjson.overrides = this.overrides;
 
     this.originalPjson.jspm = pjson;
   }
--- a/lib/config/package.js
+++ b/lib/config/package.js
@@ -167,10 +167,6 @@ PackageJSON.prototype.write = function() {
   if (this.packages != defaults.packages)
     directories.packages = this.packages;
 
-  console.log('---------------------------------------------')
-  console.log('this', this)
-  console.log('---------------------------------------------')
-
   if (hasProperties(directories)) {
     for (var d in directories)
       directories[d] = winPath(path.relative(this.dir, directories[d]));
--- a/lib/install.js
+++ b/lib/install.js
@@ -282,7 +282,15 @@ function install(name, target, options, seen) {
       if (options.inject)
         return pkg.inject(resolution, downloadDeps);
 
-      return pkg.download(resolution, options.override, options.unlink, downloadDeps);
+      var override;
+
+      try {
+        override = options.override || config.pjson.originalPjson.jspm.overrides[name];
+      } catch(e) {}
+
+      // TODO: compound the options.override onto the package.json override?
+
+      return pkg.download(resolution, override, options.unlink, downloadDeps);
     })
     .then(function(fresh) {
       resolution.fresh = fresh;


trusktr:
Alright, there's the working hack, for effect.

trusktr:
As with https://github.com/jspm/jspm-cli/pull/372, any considerations to consider, or patterns to follow?

guybedford:
We can certainly consider this.

Overrides do compound all the way through I think?

lib/config/package.js
```diff
@@ -213,6 +213,12 @@ PackageJSON.prototype.write = function() {
       if (this.originalPjson[p] === pjson[p])
         delete pjson[p];
     }
+
+    // keep original overrides
+    try {
+      pjson.overrides = this.originalPjson.jspm.overrides;
```
guybedford: We should be able to do this without `try - catch`.

lib/install.js
```diff
@@ -282,7 +282,15 @@ function install(name, target, options, seen) {
       if (options.inject)
         return pkg.inject(resolution, downloadDeps);
 
-      return pkg.download(resolution, options.override, options.unlink, downloadDeps);
+      var override;
+
+      try {
+        override = options.override || config.pjson.originalPjson.jspm.overrides[name];
+      } catch(e) {}
```
guybedford: Again, remove the `try - catch`.

lib/install.js
```diff
@@ -282,7 +282,15 @@ function install(name, target, options, seen) {
...
+
+      // TODO: compound the options.override onto the package.json override?
```
guybedford: I don't think this is necessary?

lib/config/package.js
guybedford: Follow the same pattern as in https://github.com/jspm/jspm-cli/blob/master/lib/config/package.js#L95 so that this becomes `config.pjson.overrides` when accessed from install.js.

trusktr: Ok cool, I'll come back to this tomorrow. Time for ZZzzzzz :D

trusktr:
@guybedford changed try-catch to a conditional.

lib/install.js
trusktr: Thought I'd ask about it since you're doing that with the registry overrides right?

trusktr:
@guybedford Alright, there we go, using the correct pattern.

lib/install.js
trusktr: I'm fine with the command line override simply replacing the package.json override. We could always come back to this later if necessary.

guybedford:
I don't think we should go down this road as it is against modularity. Rather I would like to focus on improving workflows around generating and sharing overrides as in https://github.com/jspm/jspm-cli/issues/336.

trusktr:
>  it is against modularity

What do you mean?

guybedford:
Having overrides is equivalent to forking a project. Forks need to be resolved otherwise that is against shared modularity.

trusktr:
But if I'm overriding locally, why does it matter? My local override is for *application* development on my end only, not for anyone else, not like package development.

guybedford:
Because if you have a shared dependency within your application tree (say lodash or jquery) that is used by other things, and you've changed the override, those other things will break.

trusktr:
In that case, can jspm just fork, leaving the note in the output?

guybedford:
@trusktr so the idea is that it is always better to share the override in the jspm registry and let others benefit from your config. Over time, this makes jspm support more and more libraries, and getting started with jspm becomes easier.

trusktr:
Totally, 100% agree with that. But things are not perfectly linear. For example, bootstrap in the jspm registry doesn't auto-load css, and I want it to, so a simple override on my side makes it happen.

guybedford:
Bootstrap in jspm registry does auto-load css.

trusktr:
Ok, maybe it does now, but not all packages will. No one wants to wait for a registry PR. They just want to get rolling in like 2 minutes.

trusktr:
I'll just wait to see how the override publish workflow will be. :)

douglasduteil:
:+1:  I have to confess that I'm overriding heavily locally without sharing it...
I end up making a gulp helper function to do it...

```js
function overwriteJspmInstall(moduleName, config, cb) {
  return spawn('sh', [
    '-c', 'jspm install ' + moduleName +
    ' -o \'' + JSON.stringify(config) + '\''
  ], {
    cwd: 'src',
    stdio: 'inherit'
  })
  .on('exit', cb)
}
```

The thing is that it's really useful when you want to use an 
unminified version a dependency like for [angular](https://github.com/jspm/registry/blob/b318ce17b7a21c8cd74a440ba82b508ab21b08ae/package-overrides/github/angular/bower-angular%401.2.1.json) : 

```js
gulp.task('jspm:angular', function (cb) {
  overwriteJspmInstall(
    'angular',
    {
      "main": "angular", // unminified here !
      "format": "global",
      "shim": {
        "angular": { // unminified here
          "exports": "angular"
        }
      }
    },
    cb
  );
});
```

Or to add a dependency between two dependencies for specific files like to define the external css of [angular-ui-select](https://github.com/angular-ui/ui-select) as the css of the [registered](https://github.com/jspm/registry/blob/d4f165e01f739856a4fb1a434b366e0ae5955154/package-overrides/github/ivaynberg/select2@3.4.5.json) [select2](https://github.com/select2/select2): 

```js
/**
 * Define the css dependencies of angular-ui-select
 */
gulp.task('jspm:angular-ui-select', ['jspm:select2'], function (cb) {
  overwriteJspmInstall(
    'angular-ui-select=github:angular-ui/ui-select',
    {
      "main": 'select',
      "format": 'global',
      "directories": {
        "lib": 'dist'
      },
      "dependencies": {
        "css": "*"
      },
      "shim": {
        "select": {
          "deps": [
          './select.min.css!',
          'select2' // say hello to a new dependence here
          ]
        }
      }
    },
    cb
  );

});

gulp.task('jspm:select2', function (cb) {
  overwriteJspmInstall(
    'select2',
    { // only care about the css here
      "main": 'select2.css!', // change the main file here
      "dependencies": {
        "css": "*"
      },
      "shim": {} // force ignore default shiming here
    },
    cb
  );
});
```

---

Same as @trusktr I'm waiting to see how  the override publish workflow will be @guybedford .

guybedford:
@douglasduteil both of the above are things that would be great to have in the main registry.

ocombe:
Hey guys, so I wanted to override a lib because I need to use an old version and it doesn't have a bower.json (it was added later).
How can I do that since there is no point to add this to the jspm registry (the current version has a bower.json)... ?
I understand your point @guybedford that we should share our config with others, but there are always gonna be edge cases, and we should be able to define an override.
Is there any chance that you'd reconsider and merge this PR ?

guybedford:
@ocombe yes this is still on track, we're working to make this area much easier. Not rushing into implementation though, so will be a month or so unfortunately. Thanks for bearing with!

jpdesigndev:
@guybedford 

My vote lies with @trusktr and others. Manual, persistent overrides are important for application development. Landing patches in large projects like angular-material, angular, famous for handling those packages in an application specific way is hard and often doesn't make much sense.

PR #372, opened 2015-01-08
Ability to specify a postinstall script in package.json's jspm.scripts.

trusktr:
For effect, it works. This one will conflict with https://github.com/jspm/jspm-cli/pull/370. One would have to be rebased after the other gets merged.

 Any considerations to consider or patterns I need to use?

### DIFF
--- a/lib/config/package.js
+++ b/lib/config/package.js
@@ -213,6 +213,12 @@ PackageJSON.prototype.write = function() {
       if (this.originalPjson[p] === pjson[p])
         delete pjson[p];
     }
+
+    // keep original scripts
+    try {
+      pjson.scripts = this.originalPjson.jspm.scripts;
+    } catch(e) {}
+
     this.originalPjson.jspm = pjson;
   }
 
--- a/lib/install.js
+++ b/lib/install.js
@@ -7,6 +7,7 @@ var PackageName = require('./config/package-name');
 var ui = require('./ui');
 var path = require('path');
 var link = require('./link');
+var exec = require('child_process').exec;
 
 var rimraf = require('rimraf');
 
@@ -295,6 +296,18 @@ function install(name, target, options, seen) {
     .then(function(depMap) {
       if (!options.parent)
         logInstall(name, target, resolution, options);
+    })
+
+    // run the postinstall script, if any.
+    .then(function() {
+      var scripts = {};
+
+      try {
+        scripts = config.pjson.originalPjson.jspm.scripts || {};
+      } catch(e) {}
+
+      if (scripts.postinstall)
+        return asp(exec)(scripts.postinstall);
     });
   });
 


guybedford:
I really do not want to implement post-install scripts. Can you describe what use cases we need this for?

trusktr:
Well, since my library doesn't have an entry point, I just want all the files in the root after install, so

```js
{
  "name": "infamous",
  // ...
  "jspm": {
    "scripts": {
      "postinstall": "mv src/* ./"
    }
  }
}
```

guybedford:
This is what `directories.lib` is for.

trusktr:
True, and it works fine for github repos. What about npm repos?

guybedford:
So my argument is that we're aiming simply to match npm behaviour exactly in npm. If you were just using npm, you would have the same problem. So it is an npm problem not a jspm problem.

trusktr:
But in npm I can solve this problem with a `postinstall` script that moves everything into the root. :D

In webpack, the problem is solved like this:

```js
// webpack.config.js
var webpack = require('webpack')
module.exports = {
    entry: "./src/app.js",
    output: {
        path: './public',
        filename: "app.js"
    },
    module: {
        loaders: [
            { test: /\.css$/, loader: "style!css" }
        ]
    },
    resolve: {
        alias: {
            famous: 'famous/src' // right here, this takes the src folder out of the way.
        }
    }
}
```

trusktr:
Wait, isn't there something like this for jspm? A map?

trusktr:
Something like 

```
{
  "map": {
    "infamous": "infamous/src"
  }
}
```

in the parent package?

guybedford:
@trusktr ok that makes sense! So if we supported npm postinstall scripts that would cover this.

Note though another way to do this would be as a prepublish script that did the same thing, with a postpublish script that cleaned it up again. That would then work with jspm and npm.

You could also do a map like:

```json
{
  "map": {
    "./mainone": "./src/mainone"
  }
}
```

To generate these.

guybedford:
@trusktr I appreciate allowing `directories.lib` here could be an option. What we could do is something like https://github.com/jspm/npm/issues/35.

trusktr:
Where do the results of the `map` appear? WOuld they appear in `config.js`?

guybedford:
* Relative local maps (like the above) create forwarding modules like we do for the main entry point.
* All other maps replace into require statements of that package.

trusktr:
What about perhaps creating a new alias feature similar to webpack's?

guybedford:
I would rather integrate support for npm postinstall scripts in https://github.com/jspm/npm/issues/36.

trusktr:
:+1:  And what about overriding them in the `jspm` field? I can easily imagine that being needed, as it would be with my package. In npm, it compiles using 6to5, but with jspm, it'll just move files around.

trusktr:
Based on my comments in https://github.com/jspm/npm/issues/36, I think it'd be better to implement it here, so that it can be used with any endpoint, giving jspm users flexibility to choose. jspm doesn't do *exactly* what npm does (all deps in a single folder vs npm's per-package node_modules folders, for example), so jspm doesn't necessarily have to do *exactly* the same with the *scripts*, but it should be optionally be able to. I truly feel that would make it easier to use. `jspm` packages don't have the ability to modify dependencies like `npm` packages can, therefore running scripts 100% of the time will not work 100% of the time.

guybedford:
@trusktr your issue is solved by allowing `directories.lib` in a jspm override.

trusktr:
@guybedford Oh? Did you decide to let `directories.lib` work with `npm:` packages?

Here's a thought! :thought_balloon:  Maybe it shouldn't be the endpoint that determines how things like `directories.lib` and install hooks are treated. It makes sense is to consider whether jspm is being used for the server side vs for the browser.

In the browser, let directories.lib always work. On the server side, ignore directories.lib. Doesn't that seem to make lots of sense? The browser is completely different than server-side nodejs/npm, and can't be treated 100% like nodejs/npm.

guybedford:
@trusktr yes, I am surprised you forgot my admittance you were right and I was wrong so soon on this point. See https://github.com/jspm/npm/issues/35. The idea was to make `directories.lib` be supported only within the `jspm` property, but not at the base level, so we can override npm, but not have any legacy effects creeping in so we retain parity.

trusktr:
@guybedford Oh, I missed that or misunderstood. xD That sounds good.

PR #399, opened 2015-01-13
Add aliases to cli.js

grey93:
I have added the aliases to the command line, but I am unable to figure out how to add them to the help menu, should I do something like
```shell
...
jspm i -> jspm install
jspm install
...
```

or at the end just give a menu

```shell
...
aliases:
i -> install
...
```

That is the main reason I didn't create a PR in the first place

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -124,6 +124,7 @@ process.on('uncaughtException', function(err) {
     case 'update':
       var doUpdate = true;
 
+    case 'i':
     case 'install':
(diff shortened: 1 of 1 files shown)


guybedford:
@grey93 thanks. `b` should probably be for bundle.

For the help, don't worry about it for now, but we can certainly look at altering this structure. As the command list grows, it could be better to divide the help into subsections, then including the shortcuts in the command list can work fine.

grey93:
can `b` be for `bundle-sfx`, since that is the command people are more likely to use, because it directly creates a single script

guybedford:
Sure.

PR #446, opened 2015-01-29
6to5 Support

guybedford:


SekibOmazic:
@guybedford Sorry for bothering again, but I have an issue with 6to5 and `build-sxf`. When using traceur  everything works fine. With 6to5 I get following message:
```Shell
Uncaught Module angular not present.
```

event though angular is present in my build.js:
```Javascript
System.register("github:angular/bower-angular@1.3.8/angular.min", [], false, function(__require, __exports, __module) {
```

And my index.html looks like this:

```html
<!--<script src="jspm_packages/traceur-runtime.js"></script>-->
    <script src="jspm_packages/6to5-polyfill.js"></script>
    <script src="build.js"></script>
```
jspm v0.11.0

Any idea what could be wrong here?

PR #447, opened 2015-01-29
Lowres documentation

crisptrutski:
Should really get `--highres-source-maps` into `jspm help` (along with the other cli options) too

### DIFF
--- a/lib/bundle.js
+++ b/lib/bundle.js
@@ -153,7 +153,11 @@ exports.bundle = function(moduleExpression, fileName, opts) {
   })
   .then(config.save)
   .then(function() {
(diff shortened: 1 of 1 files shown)


crisptrutski:
@guybedford as far as README and wikis, where would you like to document the flags to the bundle commands? Just https://github.com/jspm/jspm-cli/wiki/Production-Workflows for now?

guybedford:
For now, the documentation can't handle quirks like this, so leaving it out.

guybedford:
(Documenting broad brush strokes is the first priority, once documentation structure is in place we can come back to these quirks, but docs littered with just quirks would suck even more)

trusktr:
But I'd rather know of quirks/features up front than scratch my head for hours wondering what I'm doing wrong. xD The more docs, the better IMHO. At least if something is intended to be permanent, then doc it. The more docs, the better IMHO.

guybedford:
Will read next time :)

PR #448, opened 2015-01-30
Add bundleSFX to API

davidblurton:
This is useful to call from build tools.


guybedford:
Thanks - we're looking at providing a SystemJS builder class instance directly through the API to allow all build API options without granular wrapping. This is being tracked at https://github.com/systemjs/builder/issues/44.

Happy to take this in the mean time though.

PR #487, opened 2015-02-08
update liftoff

tkellen:
Basically all of the new features in Liftoff are aimed at letting consumers of tools written using it write their configuration files in transpile-to-js languages transparently, or use v8flags like `--harmony` to interpret them. Since you're not using these features, no code changes are needed other than bumping the dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "jspm-github": "~0.10.0",
     "jspm-npm": "~0.18.1",
     "jspm-registry": "~0.3.2",
-    "liftoff": "^0.13.2",
+    "liftoff": "^2.0.0",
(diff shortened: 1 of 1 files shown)


guybedford:
Amazing, thanks so much for this! Just wasn't sure if the implications might be more or not, so that helps a lot.

tkellen:
My pleasure man--thanks for trying to fix the hellscape that is client-side javascript dependency management. I haven't used jspm on a project yet but I'm looking forward to trying it!

guybedford:
Thanks! Still a way to go on this stuff, but do let me know your thoughts when you try it out. Appreciated for following this through.

Issue #404, opened 2015-01-14
Wildcard bundling


guybedford:
eg `jspm bundle lib/* - lib/core`

ahmedshuhel:
When I requested you said `no`. Now @EisenbergEffect did the same and you created this issue. You better be good to me :wink:

ahmedshuhel:
@guybedford I am interested in participating on this.

guybedford:
@ahmedshuhel sure, this is quite intricate as the wildcards are at a module name level. I can describe the process potentially, but it won't be easy to pick up on.

EisenbergEffect:
It would be awesome to have this same functionality available for the `depCache` generation too...

guybedford:
Ok this is now in the master branches across the projects. Merged in https://github.com/jspm/jspm-cli/commit/58ccb1ab22a9b1827abf34793bfc79923bc4193c.

Arithmetic now supports union and single module additions / removals as well (see https://github.com/systemjs/builder/issues/113, https://github.com/systemjs/builder/issues/111).

I've also added expressions for depCache as well as suggested in https://github.com/jspm/jspm-cli/commit/0eb955969812becbd330fb0758f6f6a8ecf3b3aa.

EisenbergEffect:
Woohooo! Thanks to everyone who worked on this! :fireworks: This is a big deal for us.

martingust:
Awesome!!!!

kuraga:
Yeeeeeeeessssss!!!!!!

kristianmandrup:
Awesome guys :)

OrKoN:
That's great! :+1: :+1:  Gonna test it asap. Thanks a lot @guybedford, @ahmedshuhel and others !

PR #496, opened 2015-02-10
wildcard bundling (#404)

ahmedshuhel:
Supports glob patterns like `app/*` and `app/**/*` at the moment. All the existing arithmetic operations are supported even with glob patterns. For example `main-module +  app/**/* - folderb/*` is supported too.  
Needs refactoring a lot.


lib/bundle.js
```diff
@@ -76,6 +77,101 @@ function extractBundleName(fileName) {
   return path.relative(config.pjson.baseURL, fileName.replace(/\.js$/, '')).replace(/\\/g, '/');
 }
 
+function getModuleName (file) {
```
ahmedshuhel: @guybedford I am not completely sure about this implementation. Please take a look at it.

lib/bundle.js
```diff
@@ -103,31 +199,33 @@ exports.bundle = function(moduleExpression, fileName, opts) {
     return systemBuilder.trace(firstModule, cfg);
```
ahmedshuhel: Here first module needs to be a no glob operation at the moment as `firstmodule` is being treated specially for generating the initial `trace`. We have to start the loop here instead as we did in the next `then` to support `glob` in any position.

guybedford:
Thanks for this. I'll aim to review this weekend.

lib/bundle.js
guybedford: This doesn't seem right - can we not glob prior to the trace operation and expand the glob into the arithmetic?

guybedford: If we did this, the globbing doesn't need to alter the bundle function - it can actually be done as a `globExpression` function which converts a globbed module expression into a non-globbed module expression.

For example:
```
  module/a + lib/**/* - lib/*
```
We then glob each expression individually and expand it in place:
```
  module/a + lib/sub/path + lib/a + lib/b - lib/b - lib/a
```

We don't need to worry about arithmetic reduction though.

Not saying it has to be this way, just throwing out ideas for making it consistent. Let me know what you think.

lib/bundle.js
```diff
@@ -76,6 +77,101 @@ function extractBundleName(fileName) {
...
+  return loader.normalize(moduleName)
+    .then(function (normalized) {
+      return loader.locate({name: normalized})
+    })
+    .then(function (address) {
+
+      var normalizedAddress = path.normalize(address).replace(new RegExp('\\' + path.sep, 'g'), '/'); // glob does not like `\`
+      normalizedAddress = normalizedAddress.replace(new RegExp('^file:'), '');
+
+      var files = glob.sync(normalizedAddress, {});
```
guybedford: When we do the glob, can we add glob options to remove all globbing features we're not supporting (https://github.com/isaacs/node-glob#options). I'm just worried about things like `[` or `@` in module names causing issues.

lib/bundle.js
```diff
@@ -76,6 +77,101 @@ function extractBundleName(fileName) {
...
+  }
+
+  if(id === ''){
+    throw new Error('No matching rule defined for file: ' + "'" + file + "'");
+  }
+
+  return id;
+}
+
+function sortPathsByValue(obj) {
```
guybedford: I don't think we need this function, it's fine to just walk the paths.

lib/bundle.js
guybedford: We can simplify this implementation by walking the paths with a current match and not breaking on a match, but rather updating the match if there is a better one. The most specific match should be considered the one with the largest value of `path.split('/').length`.

Apart from that the implementation looks good at first glance. Perhaps the regular expressions can be simplified?

ahmedshuhel: @guybedford Great! that makes the `path sorter function` obsolete too which you mentioned in the other comment.

lib/bundle.js
ahmedshuhel: @guybedford This is actually what I wanted to do at first. This is simple and simple is elegant :) .  Somehow got deviated by the `normalization` step required in the `expandGlob` function initially.
   
Steps:

1. Find each glob expression in the `operations` array
2. Execute the glob. get all the module names.
3. Push the every module name in the place of the corresponding `glob exprasion` in the `operations` array with the original operator. 
4. Repeat.

This gives us the opportunity to support all the glob patterns not just `lib/**/* and lib/*`. The only problem is how the program would separate the glob pattern from the `normal module name`. Simple change in the user input can make this happen. Consider the bundle expression bellow:

  `module/a + [lib/**/*] - [lib/*] + module/b + [*(a|b|c)]`

where everything inside every `[..]` is the glob expression. This way we can support complex glob pattern like the last one and we don't have to go through any additional `normalization` steps either. Let me know your thoughts.

lib/bundle.js
ahmedshuhel: If you consider my above proposal, we don't have to be worried about it either.

lib/bundle.js
ahmedshuhel: Yes.

lib/bundle.js
guybedford: @ahmedshuhel I'd still suggest that we glob in the module name space. It is important to allow users to be able to do `github:components/jquery@1.2.3/**/*` so that they don't write `jspm_packages/github/components/jquery@1.2.3/**/*`.

The worry of which syntax to allow is still valid here. So I'd still suggest creating a list of characters to escape.

ahmedshuhel:
@guybedford consider this bundle expression:
`bundle aurelia-bootstrapper + dist/**/* + github:aurelia/history-browser@0.2.3 + github:aurelia/templating-binding@0.8.4 + github:aurelia/templating-resources@0.8.4 + github:aurelia/templating-router@0.9.2 + github:aurelia/event-aggregator@0.2.2  -i`

If I omit the version( eg `github:aurelia/history-browser`), `loader.normalize` cannot resolve the module name with version. Is it an expected behavior?

Other than that, everything works fine. I optimized/organized some `regex` as you have mentioned earlier. 

`blacklisting` unsupported glob patterns will complete the implementation. working on it. @EisenbergEffect

guybedford:
Nice work @ahmedshuhel, this is looking great. The behaviour you described is exactly as currently designed - I've created https://github.com/systemjs/systemjs/issues/356 to consider an alternative.

Please just let me know when you're ready for me to review.

ahmedshuhel:
Thank you. I will let you know as soon as I can.

ahmedshuhel:
@guybedford Could you please review the code. I have added the code to guard against the unsupported patterns and characters in bundle expression [here](https://github.com/jspm/jspm-cli/blob/075df810296a76a61a0f1ac94357a40dc21101bc/lib/bundle.js#L171) and [here](https://github.com/jspm/jspm-cli/blob/075df810296a76a61a0f1ac94357a40dc21101bc/lib/bundle.js#L142)

gishmel:
Is there something causing this to not be merged?

lib/bundle.js
```diff
@@ -20,158 +20,253 @@ var Builder = require('systemjs-builder');
...
-    var depCache = config.loader.depCache = config.loader.depCache || {};
-    for (var d in traceTree) {
-      var deps = traceTree[d].deps.map(function(dep) {
-        return traceTree[d].depMap[dep];
-      });
+    .then(function() {
+      moduleName = moduleName || config.loader.main;
+    })
+    .then(function() {
+      ui.log('info', 'Injecting the traced dependency tree for `' + moduleName + '`...');
```
guybedford: I hate to be pedantic about style, but can we keep the `.then` flush with the initial statement? I know it's not the common style, but the rest of the codebase is done this way.

lib/bundle.js
```diff
@@ -20,158 +20,253 @@ var Builder = require('systemjs-builder');
...
+    throw 'Unknown operator ' + op.operator;
+  return operatorFunction;
+}
+
 // options.inject, options.sourceMaps, options.minify
 exports.bundle = function(moduleExpression, fileName, opts) {
-  var args = moduleExpression.split(' ');
 
-  var firstModule = args[0];
+  var regex = /\?|\!|\[|\]|{|}|\|/;
```
guybedford: `!` is allowed in module names - for example `jspm bundle file.css!`. Also `{` and `}` will be supported as well for conditionals.

Perhaps rather than throwing outright, just ensure the characters are simply escaped before putting into the glob function?

lib/bundle.js
```diff
@@ -20,158 +20,253 @@ var Builder = require('systemjs-builder');
...
-  for (var i = 1; i < args.length - 1; i = i + 2) {
-    var operator = args[i];
-    var moduleName = args[i + 1];
-
-    operations.push({
-      operator: operator,
-      moduleName: moduleName
+  var seq = expandGlob(args[0])
+    .then(function(moduleNames) {
+      pushOps(moduleNames, '+');
```
guybedford: The expansion should expand the operator immediately before the current module in the arithmetic.

For example:

```
 a - lib/* -> a - lib/module - lib/dep
```

lib/bundle.js
```diff
@@ -20,158 +20,253 @@ var Builder = require('systemjs-builder');
...
+    })
+    .then(function(address) {
+      var normalizedAddress = path.normalize(address).replace(/\\/g, '/').replace(/^file:/, '');
+
+      var files = glob.sync(normalizedAddress, {
+        nobrace: true,
+        noext: true,
+        nonegate: true,
+        nocomment: true,
+        nodir: true
```
guybedford: My mistake - these will already disable the other characters fine.

lib/bundle.js
```diff
@@ -20,158 +20,253 @@ var Builder = require('systemjs-builder');
...
+    });
+  }
 }
 
 function extractBundleName(fileName) {
   return path.relative(config.pjson.baseURL, fileName.replace(/\.js$/, '')).replace(/\\/g, '/');
 }
 
+function getModuleName(file) {
+  var baseURL = config.pjson.baseURL + '/';
```
guybedford: Only add the trailing `/` if it doesn't already exist. Also for windows support we probably want this to be `path.sep`?

guybedford: Actually, sorry baseURL will never have a trailing slash already.

lib/bundle.js
```diff
@@ -20,158 +20,253 @@ var Builder = require('systemjs-builder');
...
+  var paths = systemBuilder.loader.paths;
+
+  var id = '';
+  var matchLength = 0;
+
+  for (rule in paths) {
+    var resolvedPath = path.resolve(baseURL, paths[rule]).replace(/\\/g, '/');
```
guybedford: The `paths[rule].replace(/\//g, path.sep)` probable needs to be added here as well?

lib/bundle.js
```diff
@@ -20,158 +20,253 @@ var Builder = require('systemjs-builder');
...
+    var pathDir = path.dirname(resolvedPath);
+    var fileDir = path.dirname(file);
+    var pathLength = pathDir.split('/').length;
```
guybedford: Perhaps remove the `.replace(/\\/g. '/')` above and use `.split(path.sep)` here?

lib/bundle.js
```diff
@@ -20,158 +20,253 @@ var Builder = require('systemjs-builder');
...
+function expandGlob(moduleName) {
+  var loader = systemBuilder.loader;
+  return loader.normalize(moduleName)
+    .then(function(normalized) {
+      return loader.locate({
+        name: normalized
+      });
+    })
+    .then(function(address) {
+      var normalizedAddress = path.normalize(address).replace(/\\/g, '/').replace(/^file:/, '');
```
guybedford: Probably want `address.replace(/\//g, path.sep)`?

guybedford:
Nice! The algorithm looks exactly right now, let me know when you've had a chance to look at the changes.

Thanks for your patience on review and efforts to push this through. I'm doing a release this week - will aim to get this in.

lib/bundle.js
ahmedshuhel: I am developing on windows to make sure it's portable. I actually intentionally used `/`. `Glob` does not like `\` even on windows.

lib/bundle.js
ahmedshuhel: I think this is what I have done. I may not understand your comment though. Could you please take a look at it again. Your example should work too.

guybedford: Specifically the operator is hardcoded as `+` which leaves out `-`?

lib/bundle.js
guybedford: Why do we even need `path.normalize` here?

lib/bundle.js
ahmedshuhel: hardcoded `+` is for the `first module` in the expression. For example:

`a - lib/* -> + a - lib/module - lib/dep`
or
dist/* - lib/*  -> + dist/a + dist/b - lib/a - lib/b 

I assumed nobody will write a bundle expression like: `-dist/* + lib/*`

lib/bundle.js
ahmedshuhel: The same question I asked you in the beginning. But if there is a `map` exists in the `config.js` and user includes that mapped name in the bundle expression, `normalize` will return the actual module name, right?

ahmedshuhel:
@guybedford fixed the code formatting issues. Could you please take a look at it one more time? Thank you for all your help and support.

gishmel:
Is this getting merged this weekend? Or are you waiting till the systemjs merges in as well?

kuraga:
:+1:

gishmel:
So is there any idea when we can expect this feature?

guybedford:
I'm really hoping for next week - apologies for the delay, my Macbook was stolen, a clients server crashed and I had an operation yesterday. Getting to it honest.

gishmel:
That is terrible I am sorry and yeah just looking forward a lot to the help from the great tool you guys have made to assist me in bundling my application so I can push my first Aurelia application into the wild. Sorry for being a bit overzealous.

guybedford:
Thanks, no problem at all, it's good to show support to issues. Will update here soon.

EisenbergEffect:
@guybedford Get some rest man. Don't push yourself too hard. We will be fine. Were you able to recover your computer? Are you feeling ok?

clintwood:
@guybedford, take a week off man!

guybedford:
@ahmedshuhel thanks for your patience and drive to get this moving. Wildcard bundling is now merged, see https://github.com/jspm/jspm-cli/issues/404#issuecomment-83164223 for more details.

Thanks @gishmel @EisenbergEffect @clintwood for understanding - all back up to speed on a new machine from backups and will find out if my new ear drum is working soon enough.

Just firing up the windows VM for fixes there and doing final testing. The release cycle will follow within the day with some luck.

guybedford:
Just to update here - apologies, I should avoid trying to make release promises. It will definitely be out early next week though.

Issue #410, opened 2015-01-17
iojs support


capaj:
Using io.js 1.0.2 and Windows(will try on Ubuntu tommorow), I am unable to download any github packages. No error is shown-jspm just hangs there undefinately either on 
```
Updating registry cache...
``` 
or on 
```
Downloading github:urish/angular-moment@0.9.0
```

This happens with some github packages or NPM. I have got this problem with bootstrap, angular-moment so far and few others so far.

guybedford:
@capaj sure, I will need to check if io.js is supported! Are you saying it gets past that issue and does start downloading, only getting stuck occasionally?

capaj:
@guybedford no, always for some packages. For ther packages it always works.

BuildingJarl:
I noticed the same thing. 

Temp solution
cancel command (ctrl-c) after a couple of minutes then reenter it

![image](https://cloud.githubusercontent.com/assets/5493114/5819089/05abd4ea-a0b3-11e4-8284-4a980cfeef73.png)

konsumer:
I can confirm with io.js v1.0.3

capaj:
@BuildingJarl yes while you can try again and it will go through, for me the file which says 
```javascript
module.exports = require("github:angular/bower-material@0.6.1/angular-material");
```
is not created, hence your app will fail if you require('angular-material').
I guess this file is created only after downloading from github. When files are there already, JSPM does not create the public/jspm_packages/github/angular/bower-material@0.6.1.js file.

guybedford:
It sounds like this is an issue for downloading the release archives then? That would correspond to this code - https://github.com/jspm/github/blob/master/github.js#L355.

I haven't been able to replicate this myself yet in a test case. Could it just be slow downloads though?

Any further help isolating / debugging would be greatly appreciated.

capaj:
No way this is just slow download. I get this on 100 Mbit connection. Will
try to look at the code and isolate the culprit.

Dne čt, 22. 1. 2015 15:11 uživatel Guy Bedford <notifications@github.com>
napsal:

guybedford:
Can you confirm these are all iojs issues? iojs is not currently supported.

guybedford:
I've tested with iojs and can confirm that the GitHub release downloads are stalling. Updating request helped, leading to this request - https://github.com/jspm/github/blob/master/github.js#L493 sometimes returning a response and sometimes stalling.

[...]

PR #522, opened 2015-02-17
Fixes jspm install for io.js

jimthedev:
This uses a temporary fork of ncp that does not do version checking. There is a [pending pull request](https://github.com/AvianFlu/ncp/issues/79) with the ncp author that has not yet been accepted.

This should temporarily fix an issue that came up while working on #410 where `jspm install` fails silently and erratically when using io.js. Eventually when the ncp PR is merged we'd want to switch back to semver.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "liftoff": "^2.0.0",
     "minimatch": "^2.0.1",
     "mkdirp": "~0.5.0",
-    "ncp": "^1.0.1",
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks so much! Will try this out.

wesleycho:
This patch seems to work :) .  Thanks!

jimthedev:
@wesleycho Glad to hear it. Have fun.

PR #525, opened 2015-02-18
support for configurable line endings when writing config.js and package.json files

bennage:
The behavior of `JSON.stringify` on Windows with Node v0.10.35 is to use LF for print pretty format. 
I had to work around that.

One side benefit, is that the logic for making the files print pretty is now consolidated. :wink:


guybedford:
Thanks, looks great.

PR #539, opened 2015-02-23
ncp v2.0.0 removes need for ember-cli-ncp add in #410

### LINKED ISSUE #538: Switch back to ncp v2.0.0.
Looks like ncp resolved the issue it was having and has now been released at v2.0.0 so we can switch back from the ember fork.

jimthedev:
Fixes #538


jimthedev:
I tested this with `jspm install -f` in the testlibs directory on iojs 1.2.0.

guybedford:
Nice, thanks for following this through.

PR #547, opened 2015-02-26
Add a warning when bundle is called with no arguments

jackfranklin:
I misunderstood the instructions for bundle and ran `jspm bundle` with nothing else, and got this error:

```
err  TypeError: Cannot call method 'split' of undefined
         at Object.exports.bundle (/Users/jackfranklin/.nvm/v0.10.36/lib/node_modules/jspm/lib/bundle.js:82:31)
```

Took me a moment to figure it out so thought I'd update so it warns instead. Let me know what you think!

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -317,6 +317,10 @@ process.on('uncaughtException', function(err) {
       options.mangle = !options['no-mangle'];
       var bArgs = options.args.splice(1);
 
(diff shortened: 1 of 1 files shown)


guybedford:
Perfect, thanks for this - nice to make things more obvious.

jackfranklin:
You're welcome! Glad to give back :)

PR #554, opened 2015-02-28
added jshint linter and editorconfig

subesokun:
- Added jshint linter to `npm test`
- Added editorconfig
- Minor code refactoring as discussed in https://github.com/jspm/jspm-cli/commit/3d3fb0b47c6f8850c6cad3834e274cc2a51364a9#commitcomment-9734519
 
So far everything works as expected and I hope that I didn't introduce due to the many changes some unwanted bugs. Most of the linter warnings have been of type `==` vs `===` or missing `;` at the end of the lines. The diff shows also many modified lines due to trimming of trailing whitespaces.

Futhermore I'd recommend to active `qurly` so that each block or scope requires `{}` to avoid easy to overlook typos like:

```javascript
if (Error)
  console.log('Error');
return Error;
// If no error continue here ...
```

To avoid annoying build errors after pusing your code to the upstream because of some linter errors there are helpfull plugins/packages for Sublime and Atom to give you direct feedback while typing. It has only very little impact on the performance of the Editor so I'm always using it. For Atom I've installed `linter`and `linter-jshint`. Also I've installed the `editorconfig`package. 

As next step I wanted to test also [ESLint](http://eslint.org/) once as it uses an AST for linting which gives better results compared to JSHint which analyzes line for line though ESLint is x2-x3 times slower.

### DIFF
--- a/api.js
+++ b/api.js
@@ -14,7 +14,6 @@
  *   limitations under the License.
  */
 
-var core = require('./lib/core');
 var install = require('./lib/install');
 var bundle = require('./lib/bundle');
 var ui = require('./lib/ui');
@@ -52,7 +51,7 @@ API.locate = function(name, parentName) {
   .then(function(address) {
     return address.substr(5);
   });
-}
+};
 
 /*
  * jspm.on('log', function(type, msg) { console.log(msg); });
@@ -71,7 +70,7 @@ ui.useDefaults();
 
 API.promptDefaults = function(_useDefaults) {
   ui.useDefaults(_useDefaults);
-}
+};
 
 /*
  * Installs a library in the current folder
@@ -87,8 +86,8 @@ API.promptDefaults = function(_useDefaults) {
  *
  */
 API.install = function(name, target, options) {
-  return install.install.apply(install, arguments);
-}
+  return install.install.apply(install, [name, target, options]);
+};
 
 /* Uninstalls a library in the current folder.
  * returns a promise
@@ -98,15 +97,15 @@ API.install = function(name, target, options) {
  *
  */
 API.uninstall = function(names) {
-  return install.uninstall.apply(install, arguments); 
-}
+  return install.uninstall.apply(install, [names]);
+};
 
 API.import = function(moduleName, parentName) {
   return API.configureLoader()
   .then(function() {
     return System.import(moduleName, { name: parentName });
   });
-}
+};
 
 // takes an optional config argument to configure
 // returns the loader instance
@@ -130,7 +129,7 @@ API.configureLoader = function(cfg) {
       System.config(cfg);
     return System;
   });
-}
+};
 
 
 /*
@@ -144,16 +143,16 @@ API.configureLoader = function(cfg) {
 // options.minify
 API.bundle = function(expression, fileName, options) {
   return bundle.bundle(expression, fileName, options);
-}
+};
 
-/* 
+/*
  * Creates a distributable script file that can be used entirely on its own independent of SystemJS and jspm.
  * returns a promise
  * options.minify, options.sourceMaps
  */
 API.bundleSFX = function(moduleName, fileName, options) {
-  return bundle.bundleSFX(moduleName, fileName, options)
-}
+  return bundle.bundleSFX(moduleName, fileName, options);
+};
 
 /*
  * Returns a jspm-configured SystemJS Builder class
@@ -167,5 +166,4 @@ API.newBuilder = function() {
     systemBuilder.config(cfg);
     return systemBuilder;
   });
-}
-
+};
--- a/lib/config.js
+++ b/lib/config.js
@@ -53,7 +53,7 @@ exports.derivePackageConfig = function(pjson, override) {
     dpjson.registry = dpjson.registry || 'jspm';
 
   return dpjson;
-}
+};
 
 // package and loader configuration objects that are created
 exports.pjson = null;
@@ -65,14 +65,14 @@ exports.load = function(prompts) {
   if (loadPromise)
     return loadPromise;
 
-  return loadPromise = Promise.resolve()
+  loadPromise = Promise.resolve()
   .then(function() {
     if (!process.env.jspmConfigPath)
       return ui.confirm('Package.json file does not exist, create it?', true)
       .then(function(create) {
         if (!create)
           throw 'Operation aborted.';
-      })
+      });
   })
   .then(function() {
     config.pjsonPath = process.env.jspmConfigPath || path.resolve(process.cwd(), 'package.json');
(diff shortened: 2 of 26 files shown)


api.js
```diff
@@ -87,8 +86,8 @@ API.promptDefaults = function(_useDefaults) {
  *
  */
 API.install = function(name, target, options) {
-  return install.install.apply(install, arguments);
-}
+  return install.install.apply(install, [name, target, options]);
```
floatdrop: Why don't use `call` then?

```js
return install.install.call(install, name, target, options);
```

lib/package.js
```diff
@@ -595,9 +599,9 @@ exports.processPackage = function(pkg, dir, pjson, postload, isCDN) {
         curDeps = pjson.dependencies && Object.keys(pjson.dependencies) || [];
         return endpoint.build(pjson, dir);
       })
-      .catch(function(e) {
+      .catch(function() {
         throw 'Error building package `' + pkg.name + '`.';
```
floatdrop: Maybe concatenate error message like this?:

```js
throw 'Error building package `' + pkg.name + '`: ' + e.message;
```

floatdrop:
Sidenotes:

 * There are a lot of `startsWith` / `endsWith` / `contains` patterns in code - is it time to use [String.prototype](https://github.com/mathiasbynens/String.prototype.startsWith) polyfills?
 * In `lib/ui.js` escape-codes are hardcoded, [chalk](https://www.npmjs.com/package/chalk) can be used for this.

subesokun:
Hi @floatdrop, thank you a lot for your feedback and having a look on my PR! With this PR I just wanted to make the code lint free with minimum intrusion as possible because the changeset is already quite big. But I think in further PRs it would make defenetly sense to include these code improvements as described by you.

floatdrop:
@subesokun thanks for this PR, jshint and editorconfig is :100: must have.

api.js
guybedford: Yes a direct call would be nicer for these. Luckily there are no checks of `arguments.length` in install, as it does take variable argument numbers.

lib/config.js
```diff
@@ -65,14 +65,14 @@ exports.load = function(prompts) {
   if (loadPromise)
     return loadPromise;
 
-  return loadPromise = Promise.resolve()
```
guybedford: What's wrong with this pattern? Would brackets make this work?

lib/install.js
```diff
@@ -641,7 +637,7 @@ function loadExistingRange(name, parent, inject) {
 
         // if the package is installed but not in jspm_packages
         // then we wait on the getPackageConfig or download of the package here
-        return secondaryDepsPromises[parent] = new Promise(function(resolve, reject) {
```
guybedford: Again I really like this pattern for easily seeing the control flow.

lib/package.js
```diff
@@ -73,13 +73,14 @@ exports.locate = function(target) {
     if (locateCache[target.endpoint][target.package])
       return locateCache[target.endpoint][target.package];
 
-    return locateCache[target.endpoint][target.package] = Promise.resolve(endpoint.locate(target.package))
```
guybedford: Same

lib/package.js
guybedford: The endpoint does its own error reporting already - https://github.com/jspm/jspm-cli/blob/master/lib/endpoint.js#L161.

guybedford:
@subesokun this is looking great, thanks for taming the codebase!

guybedford:
@floatdrop using string polyfills and chalk is certainly welcome too.

lib/config.js
subesokun: Right, sorry I didn't test it with adding brackets around it. I've tested it now and the linter doesn't complain when the whole expression is sourrounded by brackets. I'll fix it and add the brackets also to the other places you've pointed out.

subesokun:
Mhh not sure why the `#246.2` build on Travis has failed. The build `#17` on my branch is passing https://travis-ci.org/subesokun/jspm-cli/builds/52623096

floatdrop:
@subesokun bump. Any news on this?

subesokun:
@floatdrop So far the build seems to be now fine and also there are no merge conflicts. But not sure when it will be merged by Guy. As it's quite a huge change set and as also recently the numbers of jspm installs has increased by a lot it might be difficult to find a good timing to merge this PR.

subesokun:
@guybedford Awesome :) Thank you very much!

subesokun:
@guybedford Seeing the first PR's with `jshint fix` commits remembered me that I've forgot to mention something very important.

I'd really advise to not merge any of these PRs directly into the master branch which have somewhere in their history a `jshint fix` commit. The reason therefor is that this commit just shows that the author didn't really test his code before pushing it into his fork but most importantly it will pollute the git history. Furthermore such commits falsifies the contributors statistic on Github as then people will appear as top contributors though half of their commits might have been jshint fixes.

In the end there are two options to handle such PRs:
* Reject them and ask the contributor to issue a PR with a clean history
* Manually merge the PR and perform a git rebase before pushing them into the master

Rejecting PRs seems to be the easier solution but this might end often in endless discussions and important PRs might get stuck because the contributor doesn't want to rework them or doesn't rework them in the way you want to have them. Also the atmosphere might suffer in worst case.

Actually I'm prefering the git rebase method as then you can squash and reword commits in the way you want to have them but still give the credits to the original contributor. E.g. via rebase you could squash the commits 

```
added some new feature
jshint fix
jshint fix
```

into a single  `added some new feature` commit which is perfectly valid. The git commands required for this would be

```bash
# Add the git remote to fetch the PR
git remote add subesokun git@github.com:subesokun/jspm-cli.git
git fetch subesokun some-pr-branch

# Merge PR
(git checkout master)
git merge subesokun/some-pr-branch --no-ff     # --no-ff is important here

# open an interactive dialog in which you can tell git which commits it
# shall squash, reword, ... also you can change there the order of
# the commits if wished.
git rebase -i master

# Finally push everything
(npm test)
git push origin master
```

The downside of this is of course that it will casue you more work but if has the benefit that you can control the git history to some point. The reason why I said that `--no-ff` is important is because of traceability. In some cases ff merges are possible which then skips creating the merge commit. But if this merge commit is missing you actually can't tell any longer at which point exactly a PR was merged. Some dislike this merge commits but I personally see a big value in them for the above mentioned reason.

Also with such manual merge you can avoid PR histories like 

```
added feature
fix
another fix
rebase
merge fix
some other fix
improved feature
fix for improved feature
...
```
by squashing them into a simple `added feature` commit of if wished into two commits say `added feature` and `improved feature`.

guybedford:
@subesokun are you referring to a specific commit here? Can you let me know which one. Yes ideally we should squash before merge, but I'm not too worried about it.

PR #555, opened 2015-02-28
Prefix private fields with __

floatdrop:
Since [naming conventions](https://docs.npmjs.com/files/package.json) on npm packages says, that `Any name with non-url-safe characters will be rejected. Also, it can't start with a dot or an underscore.` it would be logical to prefix private fields with `_` so they will not be filtered by `extractObj`.


guybedford:
This is really great, thanks for sorting this out!

PR #558, opened 2015-03-01
Fix EINVAL rename error

floatdrop:
This happens, when `subDir` is equal `.` - seems like node does not handle path with dot on end well.


guybedford:
Thanks. It seems we've broken in io.js for some reason, will look into this soon.

floatdrop:
@guybedford I think tests, that downloading something should be mocked, because reasons :dog: Will look at this test soon.

guybedford:
Yes in theory we should have a fully-mocked endpoint to run all the package management tests against. But unfortunately, priorities...

PR #565, opened 2015-03-04
added removal of empty directories after after removal functions

Aaike:
This update cleans empty directories after the removal functions are done.


guybedford:
Thanks for posting this so quickly. I'm not so sure about the `remove-empty-directories` dependency though just for this? We already have a full file listing from the glob, so it should be possible to use this in the code.

Aaike:
oh ok sure, ill try and get rid of it

guybedford:
Thanks - because at https://github.com/jspm/jspm-cli/blob/master/lib/build.js#L119 we have `allFiles` which is a list of all files, and `removeFiles` which lists all files being removed, we thus know what the empty directories will be so can add a stage after removing files which removes directories I think?

Aaike:
i had to do another glob at the end to get remaining files instead of useing the `allFiles` or `removefiles`. but works great now.

Aaike:
sry about all that... tried a couple of things ... it really should be fine now, but Travis keeps complaining

guybedford:
I was hoping to avoid accessing the file system again. I would have thought something like the following would work:

```javascript
// remove files
Promise.all(removeFiles.map(...))
// remove folders
.then(function() {
  // create object listing directories by file counts
  var dirFileCount = {};
  allFiles.forEach(function(file) {
    var dir = file.substr(0, file.lastIndexOf('/'));
    dirFileCount[dir] = dirFileCount[dir] + 1 || 1;
  });
  // decrement file counts from removeFiles
  removeFiles.forEach(function(file) {
    var dir = file.substr(0, file.lastIndexOf('/'));
    dirFileCount[dir] = dirFileCount[dir] - 1;
  });
  // now go through and populate removeDirs
  var removeDirs = [];
  Object.keys(dirFileCount).forEach(function(dir) {
    if (dirFileCount[dir] == 0)
      removeDirs.push(dir);
  });
  // finally do removal
  return Promise.all(removeDirs.map(function(dir) {
    return asp(fs.rmdir)(dir);
  });
});
```

Aaike:
directories don't have a trailing slash , so i can't use that to detect directories.
i was thinking checking for a file extension but some files have no extension like LICENSE files.
i can make the above code work i think , so without doing an extra glob.
but i really think i need to make use of the `filestatSync` to correctly detect directories.
would that be ok ?

guybedford:
Right so this method will only remove empty directories that weren't already empty. And if there was a directory that was empty, this would cause issues if it was within the remove path.

I'm not so sure the first point above is an issue though? Perhaps we just need to tackle the exception case described in the second point?

Aaike:
pfff... how did you just pull that code out of your head ? took me like 15 minutes really figure out what was going on and what you were trying to do :) after analyzing it it just needed some fixes , works fine now ! nice one :)

Aaike:
so does the code look ok to you now ? after another PR after this one :)

guybedford:
Perhaps test it against a case where `ignore` is a directory, which contains an empty directory? I'm still worried about it throwing here.

guybedford:
The fix is most likely to detect the EISDIR in https://github.com/jspm/jspm-cli/pull/565/files#diff-26b681bd4dac10f25d5293c7595c435bR128 and then do a directory removal at that stage.

guybedford:
Also if you can rebase and make sure the jshint is passing in Travis that would be great for a merge.

Sorry to keep pushing this back.

Aaike:
omg... i have no idea what's happening here :) i will close this and try to submit a clean PR

PR #567, opened 2015-03-06
Removal of empty directories after installation

Aaike:
you wer're right in the last PR , it was throwing an error when an empty directory had an empty directory in it in case `ignore` was used. Please have a look at my take on the solution.
i had to add files that look like directories from the removeFiles array.
This could be a problem for files that look like directories, but i've added a catch for the `EISFILE` error of the rmdir function.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -170,7 +170,7 @@ exports.filterIgnoreAndFiles = function(dir, ignore, files) {
       // do removal
       return Promise.all(removeDirs.map(function(dir) {
         return asp(fs.rmdir)(dir).catch(function(e) {
-          if (e.code === 'EPERM' || e.code === 'EISFILE' || e.code === 'ENOENT')
+          if (e.code === 'EPERM' || e.code === 'EISFILE' || e.code === 'ENOENT' || e.code === 'ENOTEMPTY')
             return;
           throw e;
         });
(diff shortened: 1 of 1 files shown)


lib/build.js
```diff
@@ -124,7 +124,59 @@ exports.filterIgnoreAndFiles = function(dir, ignore, files) {
...
+      var removeDirs = [];
+      Object.keys(dirFileCount).forEach(function(rdir) {
+        if (rdir!=='' && dirFileCount[rdir] === 0)
+          removeDirs.push(dir+'/'+rdir);
+      });
+
+      //add files from the removeFiles array that look like directories (do not have file extension)
+      //this is needed to pickup empty directories inside empty directories
+      removeFiles.forEach(function(file) {
+        var dotpos = file.lastIndexOf('.');
```
guybedford: This check isn't enough because files don't need to have extensions.

Rather I was suggesting explicitly detecting `EISDIR` in https://github.com/jspm/jspm-cli/pull/567/files#diff-26b681bd4dac10f25d5293c7595c435bR123 and then running an rmdir up there which would avoid the issue.

Sorry to be so pedantic about this, but it needs to be comprehensive. Happy to polish it myself to when I can get to it.

Aaike: yes that would be great.
i know what you mean now but i'm not sure how to tackle it.
i'm not in a rush or anything with this, so whenever you have time ;)

Issue #345, opened 2015-01-01
apply lib, files and ignore before endpoint build


guybedford:

guybedford:
Will allow link to support files and ignore, making it quicker.

andreasgrimm:
with regards to jspm linking speed, is this one still scoped to be in the v0.16 release?

guybedford:
No, this is not currently being prioritised.

guybedford:
This is covered by jspm 0.17 now.

joeldenning:
When is 0.17 planned for release? This issue has caused me a lot of waiting time for linking. To illustrate, here's an example of the type of bash I've been writing as a workaround :D
```bash
npm run build && rm -rf ~/joel-cache && mkdir ~/joel-cache && mv node_modules ~/joel-cache && npm install --production && 
jspm link npm:engagements-menu@latest -y && cd ../end-user-forms-ui && jspm install npm:engagements-menu@latest --link && 
cd ../engagements-menu && rm -r node_modules && mv ~/joel-cache/node_modules .
```

guybedford:
The goal is to get an alpha out this month, for a stable release next month, but we will have to see how things work out.

PR #568, opened 2015-03-06
fix for jspm link when a node_modules directory is present

Aaike:
when i link a local package that has a node_modules in it with a lot of files, the link process will often timeout and fail.
also it is not needed to copy the node_modules folder for each linked package as this can be huge.
on that note i also excluded the .git directory which is also not needed in the linked package. (maybe it should also ignore jspm_packages ?)


Aaike:
darn that Travis!  i'm so used to double quotes ... ;)

guybedford:
Thanks. The root fix for this is actually https://github.com/jspm/jspm-cli/issues/345.

PR #571, opened 2015-03-06
Use startsWith and endsWith from es6-shim

floatdrop:
Since upcoming releases of node will support this methods we could use them now with shim and remove shim later.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "author": "",
   "dependencies": {
     "chalk": "^1.0.0",
+    "core-js": "^0.7.0",
     "glob": "~4.3.5",
     "graceful-fs": "^3.0.5",
     "jspm-github": "~0.10.4",
--- a/cli.js
+++ b/cli.js
@@ -13,6 +13,8 @@
  *   See the License for the specific language governing permissions and
  *   limitations under the License.
  */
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -5,6 +5,7 @@
   "main": "./api.js",
   "author": "",
   "dependencies": {
+    "es6-shim": "^0.27.1",
```
guybedford: Can we rather use core-js here do you think? Since it is already in the dependency chain through Babel.

api.js
```diff
@@ -14,6 +14,8 @@
  *   limitations under the License.
  */
 
+require('es6-shim');
```
guybedford: This probably needs to be in https://github.com/jspm/jspm-cli/blob/master/cli.js as well? Or every file?

guybedford:
Sure, happy to go this route.

package.json
floatdrop: Didn't know about it, looks nicer with partial require.

api.js
floatdrop: My bad. I should have put them in every file, that needs shims.

floatdrop:
Rebased and squashed.

guybedford:
Awesome, thanks!

PR #572, opened 2015-03-07
Use strip-bom for removing byte order mark

floatdrop:
Clean up code a little with [strip-bom](https://www.npmjs.com/package/strip-bom).

### DIFF
--- a/lib/common.js
+++ b/lib/common.js
@@ -18,6 +18,7 @@ require('core-js/es6/string');
 var asp = require('rsvp').denodeify;
 var fs = require('graceful-fs');
 var config = require('./config');
+var stripBom = require('strip-bom');
 
 // the opposite of extend
 // useful for setting default config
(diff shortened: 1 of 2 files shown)


lib/common.js
```diff
@@ -68,18 +69,9 @@ exports.hasProperties = function(obj) {
...
-  .then(function(pjson) {
-    pjson = pjson.toString();
-    // remove any byte order mark
-    if (pjson.startsWith('\uFEFF')) // TODO: https://www.npmjs.com/package/strip-bom
-      pjson = pjson.substr(1);
-    try {
-      return JSON.parse(pjson);
-    }
-    catch(e) {
-      throw 'Error parsing package.json file ' + file;
```
guybedford: This shouldn't be removed.

lib/common.js
```diff
@@ -68,18 +69,9 @@ exports.hasProperties = function(obj) {
 
 exports.readJSON = function(file) {
   return asp(fs.readFile)(file)
-  .then(function(pjson) {
-    pjson = pjson.toString();
-    // remove any byte order mark
-    if (pjson.startsWith('\uFEFF')) // TODO: https://www.npmjs.com/package/strip-bom
-      pjson = pjson.substr(1);
```
guybedford: Why change this?

lib/common.js
floatdrop: Yeah, I've added https://github.com/jspm/jspm-cli/pull/572/files#diff-1ffb5a9090758229a335b02a477425c7R78 just after you posted comment.

lib/common.js
floatdrop: Just thought, if there is module to do this - why don't use it? It have tests and [works with `Buffer`](https://github.com/sindresorhus/strip-bom/blob/master/index.js#L9-L12) which should be faster a little.

guybedford: I guess I see extra dependency as a cost not a saving. I completely agree for modules that have tests and larger feature sets, but for a one liner I'd rather stick with this.

floatdrop: In some cases this is fair point, but ofter modules with large feature set have huge number of submodules. Besides this extra dependency is loaded once (and it is tiny), when jspm is installed and not affecting bundled package. But this is just my thoughts.

guybedford:
Thanks - will stick with the raw code here for now.

floatdrop:
No problems, sended pr to remove todo from code. Thanks for awesome work you have done!

PR #574, opened 2015-03-07
added npm version and npm downloads badges

subesokun:
It's quite impressive that jspm has already 14k downloads per month :)

A preview of the badges you can find here https://github.com/subesokun/jspm-cli/tree/chore-badges


guybedford:
Sure, thanks!

PR #587, opened 2015-03-15
Update SystemJS to 0.15.0

cmaher:
jspm was installing incompatible versions of SystemJs and Babel +
babel-runtime, resulting in broken files.

See systemjs/systemjs#374 and babel/babel#1022 for the conversation
chain. See also https://github.com/cmaher/systemjs-babel-runtime, which breaks with the latest
version of jspm and is fixed with this commit.

I'm not sure what tests to update with this... are there sample projects that you test against?

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -37,7 +37,7 @@ var core = module.exports;
 // we always download the latest semver compatible version
 var lVersions = {
   esml: '^0.14.0',
-  system: '^0.14.0',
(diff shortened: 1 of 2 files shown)


lookfirst:
:+1:

cmaher:
@guybedford would you instead prefer a PR that lets the user choose which library versions to use?  Right now it seems that the only available options are what's coded and edge.

PR #594, opened 2015-03-18
added jspm cache clean command

subesokun:
Added command to clean the (global) jspm cache. The implemented strategy is a bit brute-force as it searches for all folders ending with the suffix `-cache` within `~/.jspm` folder but this ensures that really all endpoint cache folders are getting removed. Futhermore the `loader-files` and `packages` folder are getting removed when cleaning the cache. So far it skips the `registry` folder as it might point to a Git repo with local modifications that are not safe to delete.

I was not sure where I could put this logic so I've created a new file `lib/cache.js` which handles these clean cache operation.


subesokun:
It needs to be also considere if it makes sense to delete the `linked` folder

guybedford:
Thanks for contributing as always. I'm just trying to understand the use case here - the package cache is effectively cleared when we run `jspm install -f` so do we need to be clearing this?

As for the loader files, what specifically made you need to add this? Was `jspm dl-loader --edge` being cached? If so we should ensure it isn't.

subesokun:
Sure, let me explain the use case. Effectively a `jspm install -f` and `jspm cache clean & jspm install` will have same effect thought `jspm cache clean` has the side effect that the `~/.jspm` folder size gets reduced. Actually my main intent was to put back the `~/.jspm` folder into some initial state just like after a fresh installation. Just the config files should be kept.

After playing around with several endpoints and experimenting with many packages the `~/.jspm` folder looked quite messed up and there was no official way yet to clean it so I thought it might be a good idea to add this `jspm cache clean` command.

Regarding the `load-files` directory. Right actually it's kind of configuration and shouldn't be deleted. I'll remove that.

guybedford:
@subesokun ok I see what you mean about it being valuable to save disk space and being useful.

Ok sure, do this feature then! Will post some implementation notes now.

lib/cache.js
```diff
@@ -0,0 +1,42 @@
...
+
+var config = require('./config');
+var path = require('path');
+var fs = require('graceful-fs');
+var rimraf = require('rimraf');
+
+exports.clean = function () {
+  var jspmDir = path.resolve(config.HOME, '.jspm');
+  var loaderFilesCacheDir = path.join(jspmDir, 'loader-files');
+  var packagesCacheDir = path.join(jspmDir, 'packages');
```
guybedford: We should also add any `jspm-*` folders as well. These correspond to the endpoint cache folders which are also useful to clear.

cli.js
```diff
@@ -23,6 +23,7 @@ var core = require('./lib/core');
 var bundle = require('./lib/bundle');
 var endpoint = require('./lib/endpoint');
 var install = require('./lib/install');
+var cache = require('./lib/cache');
```
guybedford: Rather than having a dedicated module for `cache`, lets put this in `core`. I don't think we'll have other cache functions? Also `cache` is not something to think of as a single interface as well.

cli.js
```diff
@@ -507,6 +510,11 @@ process.on('uncaughtException', function(err) {
       globalConfig.set(property, value);
       break;
 
+    case 'cache':
```
guybedford: In line with the above, perhaps lets make it `jspm cache-clear`, and `jspm cc` for short?

subesokun:
Great, thanks for your feedback :) 

As npm and bower has the same `cache clean` command implemented I've sticked to that as well but of course I can change it to `cache-clear`. The shortcut `cc` sounds handy :)

http://bower.io/docs/api/#cache
https://docs.npmjs.com/cli/cache

Regarding the `lib/cache.js` file. Initially I thought also that there might be only the `clean` function  which can be part of the `core.js` but after reading the API spec of npm and bower I'm wondering if it would make sense to add also at some later point these cache inspect functions like

```bash
jspm cache ls  // or cache-ls
jspm cache ls react 
```

`ls react` could print the dependency tree for that package or show other informations like which versions of react has been cached already. If we add such functionality it would make sense to keep the `lib/cache.js` separated.

How do you think?

guybedford:
To me, `clean` sounds like polish, while `clear` is more obvious that the cache is being removed. For example we have `jspm clean` which will clean up any dependencies not used in the project jspm_packages, but not clear it.

I know this is what npm and Bower do, and that it may confuse users, but it's not a common enough command for that to be an issue, and we can always have a message to catch a `jspm cache clean` and tell the user to write `jspm cache-clear` or `jspm cc`.

The `cache ls` stuff makes no sense to me either. Once you've installed something it is cached. Why would you care if something is in the cache already? If you want to inspect a dependency that is `jspm inspect`.

subesokun:
Yes I also was just wondering about this `ls` after writing the previous lines. Actually I've never used it. Ok then we can stick to `cache-clear` and `cc` and move it to `core.js`

subesokun:
Reworked the feature as discussed and if you want I can create a new clean PR for it or you can squash it into a single commit when merging the PR.

Seems like as the master is failing the simulated merge requests fails as well.

cli.js
```diff
@@ -102,6 +102,8 @@ process.on('uncaughtException', function(err) {
       + 'jspm config <option> <setting>     Configure jspm global options\n'
       + '                                   Stored in ~/.jspm/config\n'
       + '\n'
+      + 'jspm cache clean                   Clean the jspm cache\n'
```
guybedford: Perhaps just `cache-clean`, since we no longer have sub-commands?

guybedford:
This is great! Yes please squash and note the help text, then I will merge. The tests aren't passing as I haven't released the SystemJS builder update, so not to worry.

subesokun:
Oh sorry, I've missed to change the help text. Created fixup #606  :)

Issue #395, opened 2015-01-13
Guided override prompts


guybedford:
`jspm install bootstrap -o`

could start a prompt process for the overrides:

```
(provide an empty value to skip overriding that property)
Set the main: 
Set the module format [es6/cjs/amd/global]:
Would you like to set shim configuration [y/n]?
...
```

Hendrixer:
:+1:  This would be awesome, I have been burned by the sparse docs on how overrides work and what they are. Look at the registry to figure out how to get it going. Maybe I'll do this.

guybedford:
Initial attempted implementation at https://github.com/jspm/jspm-cli/pull/602.

guybedford:
Still not 100% sure about this though. Ideally we should delegate a certain level of responsibility here to the registry implementations.

guybedford:
This would get easier via transitive overrides allowing in-progress prompts for packages based on best-suggestions rather than a massive "wizard" concept which will be too far away from the users' needs. Tracking in transitive overrides at https://github.com/jspm/jspm-cli/issues/854.

PR #602, opened 2015-03-21
Override ideas

guybedford:
Was playing around with a concept of guided overrides today to make overrides easier.

I've decided not to try and squeeze this in with the current release as I'm not sure if this makes sense.

Will leave it here for a while instead. Feedback, ideas and contributions welcome.


guybedford:
Closed for reasons described in https://github.com/jspm/jspm-cli/issues/395#issuecomment-114569483.

PR #606, opened 2015-03-22
added jspm cache-clear command

subesokun:
Added command to clear the (global) jspm cache to free disc space.

Fixup for the PR #594


guybedford:
:+1:

PR #619, opened 2015-03-25
fixed typo in registry export

subesokun:
typo caused `jspm registry export [name]` to fail with an exception.


guybedford:
Thanks!

PR #620, opened 2015-03-25
added command to set default registry

subesokun:
e.g. `jspm registry default jspm` as more intuitive alternative to `jspm config defaultRegistry jspm`


guybedford:
For now this is not really an API that is worth publicising too heavily. I would rather wait for demand, then document `jspm config defaultRegistry` first before making its own command.

subesokun:
Ok, no problem :)

PR #642, opened 2015-03-31
fix(shim-commonjs): fix to stop format being hard-set to 'global' when a...

frankwallis:
... shim is present

So according to the spec, shims are not limited to globals, but in fact specifying a shim for a commonjs module sets the format to global in the meta, which causes 'require is not defined' errors. This change only sets the format to global if the shim has exports defined.

There are also a couple of other potential issues which I have added tests for but not fixed as I would like to know what you think first:

1) When compiling a directory it globs every single js file in the directory, normally this is not important but when doing ```jspm link``` it means that it is traversing the whole of node_modules and jspm_packages which is rather slow. If the module has directories.lib set then it avoids the problem but that is not always possible.

2) Because of the globbing, it is not possible to shim files unless they have a '.js' extension. I think that the shimmed filenames (ie the keys of 'shim') should be included in the glob. In my case I would like to shim a typescript file (.ts) with some css.

### DIFF
--- a/test/build.js
+++ b/test/build.js
@@ -1,0 +1,178 @@
+/* */
+var build = require('../lib/build');
+var asp = require('rsvp').denodeify;
+var path = require('path');
+var fs = require('fs');
+
+/*
+   options.format
+   options.shim
+   options.dependencies
+   options.map
+   options.removeJSExtensions
+   options.transpile
+   options.minify
+   options.sourceURLBase
+   options.alwaysIncludeFormat
+*/
+
+suite('compileDir', function() {
+   /* copy the fixtures to 'output' dir so we can mangle them there */
+   var ncp = require('ncp').ncp;
+   ncp.limit = 16;
+   ncp(path.join(__dirname, 'fixtures/compile/'), path.join(__dirname, './outputs/compile'), function (err) {
+      if (err) return console.error(err);
+      console.log('copied compileDir fixtures to output');
+   });
+
+   test('Allows shimming of commonjs modules and when shim is defined as array', function(done) {
+      var inputDir = path.join(__dirname, 'outputs/compile/shim/cjs/');
+      var outputFile = path.join(__dirname, 'outputs/compile/shim/cjs/entry.js');
+      var options = {
+         format: undefined,
+         dependencies: undefined,
+         shim: {
+            'entry': [ 'cjs-dep' ]
+         },
+         map: undefined,
+         removeJSExtensions: undefined,
+         transpile: undefined,
+         minify: undefined,
+         sourceURLBase: inputDir,
+         alwaysIncludeFormat: undefined
+      };
+      build.compileDir(inputDir, options)
+         .then(function() {
+            return asp(fs.readFile)(outputFile, 'utf8');
+         })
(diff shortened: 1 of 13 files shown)


test/build.js
```diff
@@ -0,0 +1,178 @@
...
+            assert(source.match(/\s*"format global";\s*/));
+            assert(source.match(/\s*"exports birmingham";\s*/));
+            assert(source.match(/\s*"deps global-dep";\s*/));
+         })
+         .then(done, done);
+   });
+
+   test('Does not process files in node_modules or jspm_packages (when linking)', function(done) {
+      var inputDir = path.join(__dirname, 'outputs/compile/glob/cjs');
+      var nodeFile = path.join(__dirname, 'outputs/compile/glob/cjs/node_modules/ignoreme.js');
```
lookfirst: According to travis, it appears you forgot to add ignoreme.js to git

lookfirst:
:+1: for tests!

test/build.js
frankwallis: oh yes, my fixture file got gitignored. updated, thanks.

guybedford:
Unfortunately this change isn't backwards compatible, so won't work in its current form.

The plan is very much to support this though, by providing a "v2" package syntax in jspm and run this update in one go.

The new package config will look like:

```json
{
  "modules": {
    "some/module.js": {
      "format": "global"
    },
    "lib/*": {
      "format": "es6",
      "plugin": "jsx",
      "deps": ["react"]
    }
  }
}
```

For example. Any help reaching this point though would be incredibly welcome. I'm going to be starting to put together the roadmap and issue steps to reach this milestone.

For the other points:

1. This is a known issue, and the suggested fix is https://github.com/jspm/jspm-cli/issues/345. Let me know if you have any more questions on that there.
2. As you can see, the new syntax above does not rely on file extension assumptions.

frankwallis:
I see, well that's ok I can just require the css in the javascript for now. 

I would really like to help out - so I guess just ping me when you have put together the roadmap and you can assign me some tasks if you want. Is there anywhere to discuss the package config changes currently or will that be on the roadmap?

guybedford:
@frankwallis thanks for expressing an interest in contributions. Getting the TypeScript stuff working well would be an amazing focus, especially with regards to the new System.register output they've been working on. If you ever want to discuss this at all please let me know. I've also updated the contributors' guide at https://github.com/jspm/jspm-cli/wiki/Contributors%27-Guide.

frankwallis:
@guybedford yes, I have been thinking about that, and it is something I could take on over the next week or two. So yes, 'assign' this to me and I will make a start.

I have created an issue on plugin-typescript https://github.com/frankwallis/plugin-typescript/issues/16 to discuss the changes. Interested to know your thoughts on that. If you also are able to quickly list the places where you know changes need to be made that would be helpful, so far I can think of:

- systemjs - doing the transpilation
- jspm-cli - configuration options and compiler installation

frankwallis:
@guybedford - I may have got the wrong end of the stick here - are you imagining that TypeScript would become an integrated transpiler option like Babel/Traceur, or do you think it should remain as a separate plugin?

Issue #640, opened 2015-03-31
jspm update to read from existing ranges


guybedford:
When using `jspm update jquery/jquery`, the update command should use the package.json ranges as a reference for the subsequent install operation to ensure we're updating within package.json ranges.

PR #667, opened 2015-04-10
Read existing ranges for `jspm update x`

crisptrutski:
Refs #640 

##### Before:

```
jspm install jquery@1.7 # installs 1.7.2
jspm update jquery      # installs 2.1.3
```

##### After:

```
jspm install jquery@1.7 # installs 1.7.2
jspm update jquery      # installs 1.11.2 (which matches 1.7 in packages.json)
jspm install jquery     # installs 2.1.3
```

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -93,8 +93,14 @@ exports.install = function(targets, options) {
     if (options.link || options.quick)
       options.lock = true;
 
+    var defaultTargets = options.dev ? config.pjson.devDependencies : config.pjson.dependencies;
     if (targets === true)
-      targets = options.dev ? config.pjson.devDependencies : config.pjson.dependencies;
+      targets = defaultTargets;
+    else if (options.doUpdate)
+      Object.keys(defaultTargets).forEach(function(module) {
+        if (module in targets)
+          targets[module] = defaultTargets[module];
+      });
 
     targets = pkg.processDeps(targets, globalConfig.config.registry);
 
--- a/cli.js
+++ b/cli.js
@@ -172,6 +172,7 @@ process.on('uncaughtException', function(err) {
     case 'install':
       options = readOptions(args, ['--force', '--override', '--link', '--yes', '--lock', '--latest', '--unlink', '--quick', '--dev']);
       options.inject = inject;
+      options.doUpdate = doUpdate;
 
       args = options.args;
 


crisptrutski:
@guybedford haven't looked to see if there's a pattern for writing test for this, just used a fresh test project

guybedford:
Awesome!

It's worth checking this will work for multiple primary installs, where only one is being updated.

crisptrutski:
Multi-"wat"

```
jspm install jquery@1.7 # installs 1.7.2
jspm install lodash@1.0 # installs 1.0.2
[update package.json range for lodash from 1.0 to ^1.0 (wasn't required for jquery)]
jspm update jquery      # installs 1.11.2, lodash updated to 1.3.1
jspm update lodash      # nothing to do
```

guybedford:
The `jspm update jquery` case should probably not touch lodash?

crisptrutski:
Yeah that's definitely a bug. Fixed that, but still confused by:

```
jspm install jquery@1.7
jspm install lodash@1.0
```

Leaving `package.json` with:
```
{
  "dependencies": {
      "jquery": "github:components/jquery@^1.7",
      "lodash": "npm:lodash@1.0"
    }
}
```

guybedford:
`jspm install jquery@1.7 lodash@1.0` definitely installs the ranges for both without any `^`. Steps to reproduce?

crisptrutski:
Perhaps memory playing tricks on me, not able to reproduc. Think this is good-to-go.

guybedford:
Ok great.

lookfirst:
I really don't understand why jspm accepts illegal semver version numbers. It clearly just complicates things unnecessarily.

crisptrutski:
@lookfirst not quite sure what you mean here - this change has nothing to do with what `package.json` version constraints look like, and merely passes on their values so that `jspm update <x>` will respect them.

lib/install.js
```diff
@@ -93,8 +93,14 @@ exports.install = function(targets, options) {
     if (options.link || options.quick)
       options.lock = true;
 
+    var defaultTargets = options.dev ? config.pjson.devDependencies : config.pjson.dependencies;
     if (targets === true)
-      targets = options.dev ? config.pjson.devDependencies : config.pjson.dependencies;
+      targets = defaultTargets;
+    else if (options.doUpdate)
+      Object.keys(defaultTargets).forEach(function(module) {
+        if (module in targets)
```
crisptrutski: @guybedford looking at this again, we should probably be checking that `targets[module]` is an empty constraint.

```
jspm install jquery@1.7
jspm update jquery@1.8 # does not update
```

guybedford: @crisptrutski I think we should probably just ignore versions for the update command. Perhaps the right behaviour would be to throw.

PR #670, opened 2015-04-10
Allow shim deps to apply for any module format

guybedford:
Only when `format: global` or exports is set do we treat as a global.

Note that this requires processing all overrides in the registry to ensure that `format: global` or `exports` are set where global processing is implied.


guybedford:
Going to bundle this with the SystemJS and config update instead to keep breaking changes together.

PR #671, opened 2015-04-10
Support "stable" prereleases

crisptrutski:
@guybedford let me know if this fits your idea of how prereleases should interact with the `stable` flag

### DIFF
--- a/lib/package.js
+++ b/lib/package.js
@@ -162,10 +162,10 @@ function getVersionMatch(pkgVersion, versions) {
     var semverMatch = version.match(semver.semverRegEx);
     var valid = semverMatch && semverMatch[1] && semverMatch[2] && semverMatch[3];
     var pre = valid && semverMatch[4];
+    var stable = versions[version].stable;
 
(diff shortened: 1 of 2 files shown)


lib/package.js
```diff
@@ -162,10 +162,10 @@ function getVersionMatch(pkgVersion, versions) {
     var semverMatch = version.match(semver.semverRegEx);
     var valid = semverMatch && semverMatch[1] && semverMatch[2] && semverMatch[3];
     var pre = valid && semverMatch[4];
+    var stable = versions[version].stable;
 
-    // ignore unstable
-    // (stable is semver, without prerelease)
-    if (!valid || pre) {
+    // ignore non-semver or prerelease, unless explictly marked as stable
+    if (!stable && (!valid || pre)) {
```
guybedford: stable === false?

guybedford: Sorry I see this case isn't the point here.

guybedford:
Beautiful.

PR #673, opened 2015-04-11
Disallow version in `jspm update x@version`

crisptrutski:
Refines #667 

Alternately, you may not want to support explicit versions in `update`


guybedford:
As mentioned in the comment I think not supporting them may be simpler for now.

crisptrutski:
:+1:, throwing occurred to me also - error message can suggest using `jspm install` instead

Issue #677, opened 2015-04-12
HOME directory is incorrect on Windows.


Zoltu:
On Windows, `%USERPROFILE%` is for files that the user directly interacts with (documents, photos, videos, downloads, etc.).  Program settings go in either `%AppData%` or `%LOCALAPPDATA%` or `%ALLUSERSPROFILE%`.  More information can be found here: http://windows.microsoft.com/en-us/windows-8/what-appdata-folder

Currently, https://github.com/jspm/jspm-cli/blob/master/lib/config.js#L16 returns `%USERPROFILE%` on Windows which is incorrect.  Not only that, but on windows `.`  prefixed files are not hidden so this clutter is readily apparent to the user.

The caveat here is that unlike Linux, there are several different well defined places to put files.  JSPM appears to be operating under the Linux model where HOME is a dumping ground for everything an application wants.  See the link above for additional details on the separation and also consider `%PROGRAMFILES%` and `%ALLUSERSPROFILE%` (`ProgramData`) for machine-wide configuration and installation (the former needs admin, the latter does not but should not contain executable code for security reasons).

I'll submit a PR for the "simple" case, though I will leave it up to JSPM maintainers to figure out a migration strategy.

PR #678, opened 2015-04-12
Changes HOME on Windows to %LOCALAPPDATA%.

Zoltu:
Fixes https://github.com/jspm/jspm-cli/issues/677 but without solving the migration problem.

A more correct cross-platform solution would be to have config.js expose multiple environment variables for correct paths depending on content.

The `.` prefix on Windows doesn't serve any purpose and can be removed, but this would require a larger refactor to have OS specific directory names.

A quick search (https://github.com/jspm/jspm-cli/search?utf8=%E2%9C%93&q=config+home&type=Code) suggests that the only thing HOME is used for is the .jspm folder so perhaps config.js could instead expose the JSPM cache/settings folder (name TBD) which could then be named more appropriately on Windows vs *nix.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -13,7 +13,7 @@
  *   See the License for the specific language governing permissions and
  *   limitations under the License.
  */
-exports.HOME = process.env.HOME || process.env.USERPROFILE || process.env.HOMEPATH;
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks for this - let me add the migration scenario as well then.

Issue #627, opened 2015-03-27
github archive as dependency


u9520107:
I tried to install engine.io-client vis jspm, in their dependencies list, there is this entry:
```
"xmlhttprequest": "https://github.com/rase-/node-XMLHttpRequest/archive/a6b6f2.tar.gz"
```
Upon hitting this entry, the installation will error out with the message:
```
err  No version match found for github:rase-/node-XMLHttpRequest@a6b6f2
```
Is this kind of behavior supported?

guybedford:
The problem here isn't the support for github archives, but the fact that we don't support commit references.

u9520107:
Ic... 
Currently, if I do engine.io-client=npm:engine.io-client, I see the error.
But if i do engine.io-client=github:Automattic/engine.io-client, then it installs fine.
I presume that it's due to the package.json having a "browser" property that specifies a different source  to the xmlhttprequest package dependency. 
Not sure why the "browser" override is respected when installing from github endpoint, but not when installing from npm endpoint.

guybedford:
It's due to the fact that `jspm install github:rase-/node-XMLHttpRequest@a6b6f2` doesn't work because `a6b6f2` is a commit reference which is not supported.

On GitHub the dependency is not installed at all, because `dependencies` in the package.json on GitHub are not installed unless a registry is explicitly specified.

lookfirst:
@guybedford case for the `@` and thrown an error to indicate that it isn't supported?

guybedford:
@lookfirst good suggestion, I've created https://github.com/jspm/github/issues/41.

PR #679, opened 2015-04-13
Support for branch names that have "/" or "@" in their name

Avnerus:
Hi,
I was trying to add an override to engine.io-client, to resolve this: https://github.com/jspm/jspm-cli/issues/627

The referenced commit in that package is in fact the latest commit on a branch called ```add/ssl-support```.
However, the installation of that branch still failed, because the ```package-name.js``` parser doesn't support ```/``` in a branch name. I noticed that ```@``` is also allowed in branch names, but not allowed in repository names, so I came up with the following fix. I tested it on the command 
```jspm install npm:feathers``` 
and it seemed to work, but I'm worried it might break other type of packages. Can you please have a look?
Thanks,
/Avner


Avnerus:
I see that this does break some tests.. Not sure about this jquery version format. Any ideas?

guybedford:
Thanks for the suggestion here - I've enabled this feature in https://github.com/jspm/jspm-cli/commit/1d6ac38afef88cad65107928a90a54370951f3be. The release won't be immediate, but you're welcome to run on master.

Avnerus:
Looking good! Thank you

PR #696, opened 2015-04-20
Enable deletion of entire symlink (but not contents)

crisptrutski:
Refs #557 

@guybedford hopefully the scariness is contained. Would really like to test this before release (tested manually in dummy project at least).

Also, what are "soft links" like on windows? Can node tell them apart?

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -882,11 +882,11 @@ function clean() {
     .map(function(dir) {
       return asp(rimraf)(dir)
       .then(function() {
-        // remove the main as well
-        return asp(fs.unlink)(dir + '.js').catch(function(e) {
-          if (e.code === 'ENOENT')
-            return;
-          throw e;
+        var filename = dir + '.js';
+        return new Promise(function(resolve, reject) {
+          fs.exists(filename, resolve);
+        }).then(function(exists) {
+          if (exists) return asp(fs.unlink)(filename);
         });
       });
     }));
@@ -900,25 +900,30 @@ exports.clean = clean;
 // depthCheck returns true to keep going (files being ignored), false to add the dir to the flat list
 function readDirWithDepth(dir, depthCheck) {
   var flatDirs = [];
-  return asp(fs.readdir)(dir)
-  .catch(function(e) {
-    if (e.code === 'ENOTDIR')
-      return;
-    throw e;
+  function pushMany(items) {
+    items.forEach(function(item) { flatDirs.push(item); });
+  }
+  return asp(fs.lstat)(dir).then(function(fileInfo) {
+    if (fileInfo.isSymbolicLink())
+      return Promise.resolve([dir]);
+
+    if (!fileInfo.isDirectory())
+      return Promise.resolve([]);
+
+    return asp(fs.readdir)(dir);
   })
   .then(function(files) {
     if (!files)
       return [];
     return Promise.all(files.map(function(file) {
-      if (depthCheck(file))
-        return readDirWithDepth(path.resolve(dir, file), depthCheck)
-        .then(function(dirs) {
-          dirs.forEach(function(subDir) {
-            flatDirs.push(subDir);
-          });
-        });
-      else
-        flatDirs.push(path.resolve(dir, file));
+      var filepath = path.resolve(dir, file);
+      /* `dir` is always absolute path, so equality implies isSymbolicLink above */
+      var notSymlink = dir != file;
+      if (notSymlink && depthCheck(filepath)) {
+        return readDirWithDepth(filepath, depthCheck).then(pushMany);
+      } else {
+        flatDirs.push(filepath);
+      }
     }));
   })
   .then(function() {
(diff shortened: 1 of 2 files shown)


lib/install.js
```diff
@@ -47,6 +47,15 @@ var installing = {
...
 };
 
+/* Unlink path, only if it exists. Returns promise. */
+function safeUnlink(path) {
+  new Promise(function(resolve, reject) {
+    fs.exists(path, resolve);
+  }).then(function(exists) {
+    if (exists) fs.unlink(path);
+  });
+}
```
guybedford: Move this down to where it is used?

lib/install.js
```diff
@@ -897,28 +902,34 @@ function clean() {
...
+
 // depthCheck returns true to keep going (files being ignored), false to add the dir to the flat list
 function readDirWithDepth(dir, depthCheck) {
   var flatDirs = [];
-  return asp(fs.readdir)(dir)
-  .catch(function(e) {
-    if (e.code === 'ENOTDIR')
-      return;
-    throw e;
+  return asp(fs.lstat)(dir).then(function(fileInfo) {
```
guybedford: I'm wondering if any users might have already symlinked their `jspm_packages` folder and this might affect that?

lib/install.js
```diff
@@ -897,28 +902,34 @@ function clean() {
...
-        return readDirWithDepth(path.resolve(dir, file), depthCheck)
-        .then(function(dirs) {
-          dirs.forEach(function(subDir) {
-            flatDirs.push(subDir);
-          });
-        });
-      else
-        flatDirs.push(path.resolve(dir, file));
+      var filepath = path.resolve(dir, file);
+      if (dir != file && depthCheck(filepath)) {
```
guybedford: Why the `dir != file` check?

lib/install.js
```diff
@@ -897,28 +902,34 @@ function clean() {
...
+        return readDirWithDepth(filepath, depthCheck).then(into.bind(flatDirs));
```
guybedford: I really prefer keeping functions inline unless they have to be shared for readability.

lib/install.js
```diff
@@ -883,11 +892,7 @@ function clean() {
       return asp(rimraf)(dir)
       .then(function() {
         // remove the main as well
-        return asp(fs.unlink)(dir + '.js').catch(function(e) {
-          if (e.code === 'ENOENT')
-            return;
-          throw e;
-        });
+        return safeUnlink(dir + '.js');
```
guybedford: Again not sure I'm a fan of a common function for a single use?

lib/install.js
guybedford: Shouldn't be an issue I don't expect, and if it turns out it is we can add a flag later. Let's not worry about this for now.

crisptrutski: We're not affecting that case - replacing death by exception with death by early return.

lib/install.js
crisptrutski: Stops eternal recur from the `Promise.resolve([dir])` above

crisptrutski: Maybe deserves comment

guybedford: Does that not stop recursion of a folder with a subfolder by the same name?

lib/install.js
guybedford: Ping

crisptrutski: Moved it down to above the calling function. You mean *inside* the function?

lib/install.js
crisptrutski: `dir` should have run through resolve, but double checking

lib/install.js
```diff
@@ -897,28 +902,36 @@ function clean() {
 }
 exports.clean = clean;
 
+/* Like concat, but in place */
+function into(extras) {
+  var list = this;
+}
```
guybedford: Do we still need this?

crisptrutski: Looks important

guybedford: Well it's not being called anywhere.

lib/install.js
```diff
@@ -900,25 +900,30 @@ exports.clean = clean;
 // depthCheck returns true to keep going (files being ignored), false to add the dir to the flat list
 function readDirWithDepth(dir, depthCheck) {
   var flatDirs = [];
-  return asp(fs.readdir)(dir)
-  .catch(function(e) {
-    if (e.code === 'ENOTDIR')
-      return;
-    throw e;
+  var pushMany = function(items) {
```
guybedford: function pushMany?

lib/install.js
```diff
@@ -900,25 +900,30 @@ exports.clean = clean;
...
-        .then(function(dirs) {
-          dirs.forEach(function(subDir) {
-            flatDirs.push(subDir);
-          });
-        });
-      else
-        flatDirs.push(path.resolve(dir, file));
+      var filepath = path.resolve(dir, file);
+      /* `dir` is always absolute path, so equality implies isSymbolicLink above */
+      var notSymlink = dir != file
```
guybedford: Semicolon?

guybedford:
Looks good, happy to merge.

lib/install.js
crisptrutski: Taking RRs (rename requests)

Issue #691, opened 2015-04-18
Globally configure default transpiler


swirlycheetah:
I'd prefer it if `jspm init` defaulted to Babel rather than Traceur but I understand that's just my own personal preference.

I was more wondering whether you collect any stats and if so could those be used to to decide which transpiler to default to?

lookfirst:
Yes!

clintwood:
Maybe this could be a user setting in `~/.jspm/config`

guybedford:
Good point - this could be a global configuration option. Then if the user chooses Babel when the default global option is set to Traceur, we can ask if they want to change the global default.

crisptrutski:
@guybedford FYI there's no "Would you like this to be the new default?" step in swirly's patch, but not sure that's necessary. 

Only updates on manual input, not `--prompts` style, which could be confusing.

guybedford:
It's quite nice having the `dl-loader` not being permanent - happy to leave as-is for now.

@swirlycheetah thanks for the PR!

swirlycheetah:
Thanks folks. Yeah it's a pretty basic implementation for now, I'm happy to extend it if required in the future. There may also be some scenarios where the tests could fail if babel is set as the default so if that comes up I'll take a look.

PR #697, opened 2015-04-20
Add globally configured transpiler default

swirlycheetah:
Covers #691, defaults to Traceur to begin with but the default will switch depending on the last entered choice made by the user.

As mentioned in gitter `npm test` failed due to some jshint errors, I've fixed those up in the most basic way but not sure if that may cause regressions.

### DIFF
--- a/lib/config/loader.js
+++ b/lib/config/loader.js
@@ -26,6 +26,7 @@ var alphabetize = require('../common').alphabetize;
 var stringify = require('../common').stringify;
 var ui = require('../ui');
 var extractObj = require('./utils').extractObj;
+var globalConfig = require('../global-config');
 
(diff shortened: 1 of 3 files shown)


crisptrutski:
@swirlycheetah your lint fixes are fine - can rebase your commit out if you want, because I'd added them here too: https://github.com/jspm/jspm-cli/commit/2aa6c4d82ce326210477141a4cd101c732f74d34

crisptrutski:
Looks good to me

crisptrutski:
Merged here: https://github.com/jspm/jspm-cli/commit/dd373a720dd7163b3b8b37ba2eb6485c0cfc591a

PR #701, opened 2015-04-22
Support convention of "latest" version

crisptrutski:
Refs jspm/npm#34

@guybedford so github and npm endpoints would be updated to return things like `{versions: [...], latest: 'master'}`

Passing in `options`, rather than `latestVersion` directly, to anticipate `--edge` option.

### DIFF
--- a/lib/package.js
+++ b/lib/package.js
@@ -128,7 +128,8 @@ exports.lookup = function(pkg) {
     lookups[pkg.package] = lookup.versions;
 
     return function(version) {
-      var lookupObj = getVersionMatch(version, lookup.versions);
+      var options = {latestVersion: lookup.latest};
+      var lookupObj = getVersionMatch(version, lookup.versions, options);
       if (!lookupObj)
         return;
 
--- a/lib/package.js
+++ b/lib/package.js
@@ -128,8 +128,7 @@ exports.lookup = function(pkg) {
     lookups[pkg.package] = lookup.versions;
 
     return function(version) {
-      var options = {latestVersion: lookup.latest};
-      var lookupObj = getVersionMatch(version, lookup.versions, options);
+      var lookupObj = getVersionMatch(version, lookup.versions, {latestVersion: lookup.latest});
       if (!lookupObj)
         return;
 
@@ -175,12 +174,15 @@ function getVersionMatch(pkgVersion, versions, options) {
     var latest = options && options.latestVersion && versions[options.latestVersion];
     if (latest && latest.stable !== false) return latest;
     stableVersions.sort(compareDesc);
-    if (stableVersions[0]) return versions[stableVersions[0]];
+    if (stableVersions[0])
+      return versions[stableVersions[0]];
 
     if (latest) return latest;
     unstableVersions.sort(compareDesc);
-    if (unstableVersions[0]) return versions[unstableVersions[0]];
-  } else {
+    if (unstableVersions[0])
+      return versions[unstableVersions[0]];
+  }
+  else {
     var i, ver;
     stableVersions.sort(compareDesc);
     // find highest stable match in tags
--- a/lib/package.js
+++ b/lib/package.js
@@ -187,13 +187,13 @@ function getVersionMatch(pkgVersion, versions, options) {
     for (i = 0; i < stableVersions.length; i++) {
       ver = stableVersions[i];
       if (semver.match(pkgVersion, ver))
-        return ver;
+        return versions[ver];
     }
     unstableVersions.sort(compareDesc);
     for (i = 0; i < unstableVersions.length; i++) {
       ver = unstableVersions[i];
       if (semver.match(pkgVersion, ver))
-        return ver;
+        return versions[ver];
     }
     // finally check for an exact tag match
     if (versions[pkgVersion])
--- a/lib/package.js
+++ b/lib/package.js
@@ -140,66 +140,64 @@ exports.lookup = function(pkg) {
 };
 
 exports.getVersionMatch = getVersionMatch;
-function getVersionMatch(pkgVersion, versions) {
+function getVersionMatch(pkgVersion, versions, options) {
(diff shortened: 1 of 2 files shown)


lib/package.js
```diff
@@ -128,7 +128,8 @@ exports.lookup = function(pkg) {
     lookups[pkg.package] = lookup.versions;
 
     return function(version) {
-      var lookupObj = getVersionMatch(version, lookup.versions);
+      var options = {latestVersion: lookup.latest};
```
guybedford: Inline!

lib/package.js
```diff
@@ -140,66 +141,64 @@ exports.lookup = function(pkg) {
...
-  // next try an unstable version range match in tags then branches
-  for (var j = 0; j < versionList.length; j++) {
-    if (semver.match(pkgVersion, versionList[j]))
-      return versions[versionList[j]];
-  }
-  for (var k = 0; k < exactVersions.length; k++) {
-    if (semver.match(pkgVersion, exactVersions[k]))
-      return versions[exactVersions[k]];
+    var latest = options && options.latestVersion && versions[options.latestVersion];
+    if (latest && latest.stable !== false) return latest;
```
guybedford: newline

lib/package.js
```diff
@@ -140,66 +141,64 @@ exports.lookup = function(pkg) {
...
+    stableVersions.sort(compareDesc);
+    if (stableVersions[0]) return versions[stableVersions[0]];
```
guybedford: and again

lib/package.js
```diff
@@ -140,66 +141,64 @@ exports.lookup = function(pkg) {
...
+
+    if (latest) return latest;
+    unstableVersions.sort(compareDesc);
+    if (unstableVersions[0]) return versions[unstableVersions[0]];
+  } else {
```
guybedford: Prefer the `}\n else {` convention here.

guybedford:
Because of the GitHub issue, I'm wondering if we shouldn't just do:

* `latestVersion` installed whenever a range is not given.
* Have a separate `edgeVersion` reference.
* Ignore whether latest or edge is stable itself.
* When installing into a range, just follow normal semver rules.

Then if endpoints don't set `latestVersion` or `edgeVersion`, we default to semver orderings again, so that endpoints don't have to opt-in and we maintain backwards-compatibility.

guybedford:
The GitHub issue being https://github.com/jspm/github/pull/43#issuecomment-95160107.

crisptrutski:
That does simplify behaviour quite a bit, not sure which I prefer

* latestVersion installed whenever a range is not given.

Guessing that if neither latestVersion or range is given, we fall back to `stables[0] || unstable[0]`

* Have a separate `edgeVersion` 

This means sorting version earlier, in the endpoint, so more work in more places, more code. Not a major objection

* Ignore whether latest or edge is stable itself 

This seems redundant with first bullet point? OK

guybedford:
All the above sound right, yes. Are there cases in the current logic that would be handled better than the simplified logic above?

guybedford:
We probably do still need to take into account the stability of `latestVersion` because GitHub master should not take preference to 1.0.0. So we probably retain most of the existing logic for unstable latest handling.

crisptrutski:
@guybedford sounds like you're happy for merge?

Will extend to handle `--edge` in a separate PR

lib/package.js
```diff
@@ -140,66 +140,67 @@ exports.lookup = function(pkg) {
...
-      return versions.master;
-    return versions[versionList[0] || exactVersions[0]];
-  }
-
-  // next try an unstable version range match in tags then branches
-  for (var j = 0; j < versionList.length; j++) {
-    if (semver.match(pkgVersion, versionList[j]))
-      return versions[versionList[j]];
+    var latest = options && options.latestVersion && versions[options.latestVersion];
+    if (latest && latest.stable !== false) return latest;
```
guybedford: newline please

guybedford: Perhaps we should take into account default stability as well here - `1.0.0-beta` is not stable, even if latest, unless explicitly overriden?

lib/package.js
```diff
@@ -140,66 +140,67 @@ exports.lookup = function(pkg) {
...
+    stableVersions.sort(compareDesc);
+    if (stableVersions[0])
+      return versions[stableVersions[0]];
+
+    if (latest) return latest;
```
guybedford: newline

lib/package.js
crisptrutski: Makes sense

crisptrutski:
@guybedford this should handle the non-edge cases, while start on second PR now

lib/package.js
```diff
@@ -140,66 +140,89 @@ exports.lookup = function(pkg) {
...
     pkgVersion = decodeURIComponent(pkgVersion);
 
-  var latestPre;
-  var versionList = [];
-  var exactVersions = [];
+  var version;
+  var stableVersions = [];
+  var unstableVersions = [];
+  var exactStableVersions = [];
+  var exactUnstableVersions = [];
```
guybedford: For clarity, would it make sense to call these - `stableSemvers`, `unstableSemvers`, `stableTags`, `unstableTags`?

guybedford:
Looking good.

guybedford:
What do you mean a second PR? For which features?

lib/package.js
crisptrutski: Not sure about `tags`. `(un)stableOther` perhaps?

crisptrutski:
The `--edge` behaviour #753

lib/package.js
guybedford: Why not tags?

guybedford:
Great! :+1: pending naming.

lib/package.js
guybedford: `stableSemver`, `unstableSemvers`, `stableExact`, `unstableExact`?

PR #742, opened 2015-05-08
Added strictSSL support from global config

jschultzmn:


jschultzmn:
Accidental pull request, not ready yet

PR #746, opened 2015-05-09
Strict ssl

### LINKED ISSUE #725: strictSSL option
A global-level config option, which is added to endpoint options for each registry constructor. Then each registry implementation usage of `request` should follow it.

smastous:
Closes #725

### DIFF
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -194,9 +194,12 @@ exports.configure = function(registry) {
   }
 
   registryConfig.name = registry;
+  registryConfig.strictSSL = globalConfig.config.strictSSL;
+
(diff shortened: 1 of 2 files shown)


lib/registry.js
```diff
@@ -65,6 +65,7 @@ exports.load = function(registry) {
     }, globalConfig.config.registries[registry] || {});
 
     options.name = registry;
+    options.strictSSL = 'strictSSL' in globalConfig.config ? globalConfig.config.strictSSL : true;
```
guybedford: Surely this can just be `globalConfig.config.strictSSL` if we have the default set in global-config as above?

guybedford:
Thanks so much for this work!

lib/registry.js
smastous: Totally agree.  Fixing now.

PR #753, opened 2015-05-12
More aggressive --edge version resolution

crisptrutski:


guybedford:
Let's merge https://github.com/jspm/jspm-cli/pull/701 and rebase first?

cli.js
```diff
@@ -65,6 +65,8 @@ process.on('uncaughtException', function(err) {
       + '\n'
       + '  install react --lock             Stable install, locking existing dependencies\n'
       + '\n'
+      + '  install react --edge             Untable install\n'
```
guybedford: To save space -  `  install react=npm:react --edge   Install a package from a registry to latest unstable\n'

crisptrutski:
@guybedford there's an issue here around npm - the dist-tag is usually the latest stable version, so don't think we can prioritize it in the "--edge, no range provided" scenario.

Example:

`jspm install react@0.14`: fail, no stable
`jspm install react@0.14 --edge`: installs 0.14.0-alpha1
`jspm install react --edge`: installs 0.13.3

Moving `latest` precedence to after semver and before exact, when edge, seems appropriate.

crisptrutski:
Additionally, `jspm update` after `jspm install react@0.14 --edge` results in `err  No version match found for npm:react@0.14`. 

Seems edge installs should save the resolved version to config, and never the range

PR #757, opened 2015-05-14
JSPM Install should install devDependencies

### LINKED ISSUE #747: JSPM Install should install devDependencies.
`npm install` installs `dependencies` and `devDependencies`.  JSPM should do the same thing when `jspm install` is executed.  When I first checkout a repository, I do not want to have to execute `jspm install` and `jspm install --dev` just to get my repo into a state that I can start working on it.

crisptrutski:
Closes #747

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -351,10 +351,17 @@ function install(name, target, options, seen) {
         primaryRanges[name] = target.copy();
       // store in package.json
       if (!options.link) {
-        if (!options.dev)
+        if (name in config.pjson.dependencies)
           config.pjson.dependencies[name] = primaryRanges[name];
-        else
+        else if (name in config.pjson.devDependencies)
           config.pjson.devDependencies[name] = primaryRanges[name];
+        else if (!options.dev) {
+          config.pjson.dependencies[name] = primaryRanges[name];
+        }
+        else {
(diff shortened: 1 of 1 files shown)


guybedford:
I'd suggest testing that this doesn't result in all dependencies being saved back into the `dependencies` object.

guybedford:
I think you'll see `traceur` and `babel` have become normal dependencies when doing this.

crisptrutski:
In my testing the behaviour was as expected (existing project, installing new dependencies with and without `--dev`).

How did you test it?

guybedford:
`jspm install` with no arguments.

crisptrutski:
See it, will touch up

crisptrutski:
Came up with a solution, let me know if you're satisfied (unfortunately all four branches can occur)

crisptrutski:
Potentially problematic: can't install different versions to each group, or move dependencies between groups, using `jspm install` with this approach

guybedford:
Commit?

crisptrutski:
Network woes it seems, pushed again

guybedford:
Nice. It seems to look good to me. Let's merge, then I'll do some more testing on master for feedback.

PR #761, opened 2015-05-15
Revise precedence of `latest` when installing `edge`

crisptrutski:
Refs #753

Addresses unexpected behaviour mentioned in comment

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -228,6 +228,13 @@ function install(name, target, options, seen) {
       else
         target.setVersion(resolution.version);
     }
+    else if(options.edge && !options.parent && !options.link) {
+      // use strictest compatible semver range if installing --edge without target, or
+      // with a range that does not include the resolved version
+      if (!target.version || !semver.match(target.version, resolution.version)) {
+        target.setVersion('^' + resolution.version);
+      }
+    }
 
     // load our fork ranges to do a resolution
     return loadExistingForkRanges(resolution, name, options.parent, options.inject)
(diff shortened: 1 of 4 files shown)


lib/package.js
```diff
@@ -185,7 +185,7 @@ function getVersionMatch(pkgVersion, versions, options) {
 
   if (!pkgVersion) {
     var latest = options && options.latestVersion && versions[options.latestVersion];
-    if (latest)
+    if (!edge && latest)
```
guybedford: :+1:

guybedford:
All looks good. Agree the `--edge` range should be replaced by the actual. This can be done by enabling this condition to apply for edge as well - https://github.com/jspm/jspm-cli/pull/761/files#diff-f6618e1cc731d58106a806b7679a7616R225.

lib/install.js
```diff
@@ -110,9 +110,10 @@ exports.install = function(targets, options) {
     else if (targets && options.doUpdate)
       Object.keys(targets).forEach(function(module) {
```
guybedford: Can we comment what this whole block is doing?

guybedford:
Better link to that code - https://github.com/jspm/jspm-cli/blob/master/lib/install.js#L224.

crisptrutski:
@guybedford took a feather from cap, but discarding user supplied ranges only when necessary (since we don't provide edge-range literals to store)

lib/install.js
```diff
@@ -227,6 +228,13 @@ function install(name, target, options, seen) {
       else
         target.setVersion(resolution.version);
     }
+    else if(options.edge && !options.parent && !options.link) {
```
guybedford: space after `else if (`

guybedford:
:+1:

PR #762, opened 2015-05-15
Avoid generating nonsensical "did you mean" suggestions

### LINKED ISSUE #760: Update has trouble with pointer packages
Confused the redirection with having a new version explicitly provided (which is not supported)

crisptrutski:
Closes #760

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -110,9 +110,10 @@ exports.install = function(targets, options) {
     else if (targets && options.doUpdate)
       Object.keys(targets).forEach(function(module) {
         var target = targets[module];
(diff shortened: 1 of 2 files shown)


guybedford:
:+1:

I'm not sure I understand your PR pattern as things seem to double-up a lot? Would it make more sense to have a branch, and the PR against that branch?

crisptrutski:
The duplicates commits were just to fixing travis (master was failing). If you mean the number of PRs per theme, that's just lack of foresight / merging too early.

Don't think we need a dev / integration branch.

crisptrutski:
Will merge this in another PR

PR #775, opened 2015-05-19
Recursive clean-up on uninstalls

### LINKED ISSUE #745: Uninstall to clear parent folder when empty
From https://github.com/jspm/github/issues/48:

> If you uninstall a GitHub repo named `x/y`, then the `y` folder is removed, but not the empty `x` folder.

crisptrutski:
Closes #745

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -869,6 +869,34 @@ function showVersions(forks) {
 }
 exports.showVersions = showVersions;
 
+/* Recursively remove directory, all those above it, if they are empty.
+ * Takes optional `stopDir` to terminate at. */
+function cascadeDelete(dir, stopDir) {
+  if (dir && dir !== stopDir) {
+    return new Promise(function(resolve) {
+      fs.exists(dir, resolve);
+    })
+    .then(function(exists) {
+      if (!exists)
+        return true;
+
+      return asp(fs.readdir)(dir)
+      .then(function(paths) {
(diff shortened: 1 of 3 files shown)


lib/install.js
```diff
@@ -869,6 +869,34 @@ function showVersions(forks) {
 }
 exports.showVersions = showVersions;
 
+/* Recursively remove directory, all those above it, if they are empty.
+ * Takes optional `stopDir` to terminate at. */
+function cascadeDelete(dir, stopDir) {
```
guybedford: Perhaps let's put this into the common.js file?

guybedford:
:+1:

crisptrutski:
@guybedford managed to find a subtle issue where parallel deletes occur in the gap between `exists` check and `rmdir`. 

Might not even be worth calling `exists` anyway now, or `readdir`, and just use the error codes. Less code

guybedford:
Well caught, yeah async needs a lot of thought. Error codes via a single access is the recommended approach for this stuff anyway. Do put some thought to symlinks as well here - they shouldn't be followed.

crisptrutski:
Not sure we should worry about symlinks here - if symlinking to a shared deps folder, think that removing empty directories is still OK. 

We could check in advance if any of the parent directories are symlinks, and then skip all of them. Is that what you mean? Can't think of anything else that makes sense to me

PR #784, opened 2015-05-21
Fix #783, added --inline-source-maps in in bundle and bundle-sfx command

### LINKED ISSUE #783: Add --inline-source-map option to bundle commands

johnraz:
Shoudl fix #783, I just made the option `--inline-source-maps` instead of `--inline-source-map`, mind the trailing *s*.

@crisptrutski , please give it a look.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -366,7 +366,7 @@ process.on('uncaughtException', function(err) {
       options.mangle = !options['no-mangle'];
       
       if (options['inline-source-maps']){
-        options.sourceMapContents = true;
+        options.sourceMaps = "inline";
       }
 
       var bArgs = options.args.splice(1);
--- a/cli.js
+++ b/cli.js
@@ -366,7 +366,7 @@ process.on('uncaughtException', function(err) {
       options.mangle = !options['no-mangle'];
       
       if (options['inline-source-maps']){
-        options.sourceMapContents = 'inline';
+        options.sourceMapContents = true;
       }
 
       var bArgs = options.args.splice(1);
--- a/cli.js
+++ b/cli.js
@@ -358,12 +358,17 @@ process.on('uncaughtException', function(err) {
       var sfxBundle = true;
 
     case 'bundle':
-      options = readOptions(args, ['--inject', '--yes', '--skip-source-maps', '--minify',  '--no-mangle', '--hires-source-maps', '--no-runtime']);
+      options = readOptions(args, ['--inject', '--yes', '--skip-source-maps', '--minify',  '--no-mangle', '--hires-source-maps', '--no-runtime', '--inline-source-maps']);
       if (options.yes)
         ui.useDefaults();
       options.sourceMaps = !options['skip-source-maps'];
       options.lowResSourceMaps = !options['hires-source-maps'];
       options.mangle = !options['no-mangle'];
+      
+      if (options['inline-source-maps']){
+        options.sourceMapContents = 'inline';
+      }
+
       var bArgs = options.args.splice(1);
 
       if (bArgs.length === 0) {


cli.js
```diff
@@ -358,12 +358,17 @@ process.on('uncaughtException', function(err) {
...
-      options = readOptions(args, ['--inject', '--yes', '--skip-source-maps', '--minify',  '--no-mangle', '--hires-source-maps', '--no-runtime']);
+      options = readOptions(args, ['--inject', '--yes', '--skip-source-maps', '--minify',  '--no-mangle', '--hires-source-maps', '--no-runtime', '--inline-source-maps']);
       if (options.yes)
         ui.useDefaults();
       options.sourceMaps = !options['skip-source-maps'];
       options.lowResSourceMaps = !options['hires-source-maps'];
       options.mangle = !options['no-mangle'];
+      
+      if (options['inline-source-maps']){
+        options.sourceMapContents = 'inline';
```
crisptrutski: This needs to be `options.sourceMaps = 'inline';`

crisptrutski:
Works for me with a slight tweak, thanks!

crisptrutski:
Unfortunately no tests for these things, but "contributions welcome" :dancer:

cli.js
johnraz: Works only with `options.sourceMapContents` for me.
Using `options.sourceMaps` raise an exception:
```
err  SyntaxError: Unexpected token i
         at Object.parse (native)
         at new SourceMapConsumer (file:some-path/jspm_packages/npm/source-map@0.1.43/lib/source-map/source-map-consumer.js:14:24)
         at fromString (file:some-path/jspm_packages/npm/clean-css@3.1.9/lib/utils/input-source-map-tracker.js:19:26)
         at InputSourceMapStore.track (file:some-path/jspm_packages/npm/clean-css@3.1.9/lib/utils/input-source-map-tracker.js:107:54)
         at Object.whenDone (file:some-path/jspm_packages/npm/clean-css@3.1.9/lib/clean.js:76:46)
         at processNext (file:some-path/jspm_packages/npm/clean-css@3.1.9/lib/imports/inliner.js:83:115)
         at importFrom (file:some-path/jspm_packages/npm/clean-css@3.1.9/lib/imports/inliner.js:65:12)
         at processNext (file:some-path/jspm_packages/npm/clean-css@3.1.9/lib/imports/inliner.js:83:74)
         at importFrom (file:some-path/jspm_packages/npm/clean-css@3.1.9/lib/imports/inliner.js:65:12)
         at inlineLocalResource (file:some-path/jspm_packages/npm/clean-css@3.1.9/lib/imports/inliner.js:228:12)
```

crisptrutski: Ultimately `options.sourceMaps` is used to determine whether the maps are inlined: https://github.com/systemjs/builder/blob/master/lib/output.js#L160-163

On my side, with your PR as it is I've tested and the sources were not inlined (and works with the change).

Is there more to the stack for that exception? Would like to replicate, but have a feeling it's related to how you're bundling CSS - which I'm not very familiar with.

johnraz: That's the complete stack trace, yes.
I have no idea where the error is coming from.

I don't do anything fancy to bundle css, its done in a separated process using lessc which doesn't interact with this. It's probably due to one of my deps.

I think we need to differentiate the line you pointed and this [one](https://github.com/systemjs/builder/blob/master/lib/output.js#L69).

The way I understand what's going on, my version of the code embeds the source in the sourcemap, while yours embeds the sourcemap in the source, right ?

all you should need to reproduce is [here](https://gist.github.com/johnraz/03fae8f6fd602f480fe1)

guybedford: This sounds like an error in the CSS plugin due to https://github.com/systemjs/plugin-css/blob/master/css-builder.js#L32 passing in the full value of `inline`.

Setting `sourceMapContents: true` seems like a viable alternative, but it should be `true` and not `inline` I think.

In terms of the API, we should probably try and converge on one or the other method though. I quite like the separate property actually to avoid issues like that.

johnraz: @guybedford the `inline` value comes from here: https://github.com/systemjs/builder/blob/master/lib/builder.js#L26

crisptrutski: @johnraz Yeah that's the right summary for the behaviours - although I'd remove some ambiguity by saying "embeds the sourcemap in the output" rather :wink: 

In JSPM we want to reserve `inline` to refer to embedding the sourcemap in the output file, and do not want to use this command line option to refer to the weaker "include sources in the source map" action. 

We should add a separate command line option for `sourceMapContents`, like `--source-map-include-contents`.

RE: the exception you're getting, I think https://github.com/systemjs/plugin-css/pull/37 should sort you out, but it has not been tested yet

crisptrutski: @johnraz tested out bundling some css with that plugin-css update - seems to work now. published version `0.11.1` - let me know if your troubles are resolved

johnraz: As stated on the related plugin-css PR, its fixed indeed ! Thanks.
I would say that we can drop the weaker option completely and stick with the current option and your comment applied (see my latest commit)

crisptrutski:
Thanks for the help discovering, fixing and testing :+1:

johnraz:
waiiit :D I just pushed a linter fix ;-)

johnraz:
Ok I was too late, I made a new PR :)

PR #792, opened 2015-05-26
Added option to not use a transpiler during init

JoshAntBrown:


guybedford:
Awesome, thank you!

JoshAntBrown:
No problem :-)

curran:
Wonderful! So happy to hear this has been implemented. Thank you for addressing this.

simonh1000:
I have the following in my config.js file and am finding that it is trying to load a file `none.js` as a result, which is causing load errors
```
System.config({
  "transpiler": "none",
```

guybedford:
@simonh1000 yes it would do that if you attempt to load ES6 modules in the loader!

CxRes:
@guybedford I am also having the `none.js` issue. Since I am using atom-electron (which uses v8 44.0 whi), I have no need for transpilation (hence `"transpiler":"none"`) but still need the package management.  Is there something wrong in my approach? Could you please suggest a workaround?

guybedford:
In the next SystemJS release this error message will be showing instead which should help here - https://github.com/systemjs/systemjs/commit/336008b488d30a4114fa616f7c8780c9915b75bb#diff-41d0147f345a3c4223cb35393ff4385fR21. Basically it will tell you which module was attempted to be loaded as ES6 that it needed to load the transpiler for.

PR #802, opened 2015-06-02
feat(unbundle): expose unbundle in the public api

mbroadst:
It may be desirable as part of an automated task (in e.g. gulp) to
be able to unbundle programmatically, just like you can currently
bundle.

### DIFF
--- a/api.js
+++ b/api.js
@@ -147,6 +147,13 @@ API.bundle = function(expression, fileName, options) {
 };
 
 /*
+ * Returns a promise for unbundling
+ */
+API.unbundle = function() {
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks for contributing to filling in the API. Unfortunately it's already there though!

mbroadst:
@guybedford ah! my bad I must be stuck on an old version :smile:

mbroadst:
@guybedford ah no maybe its part of an unreleased version? I'm not getting that api with 0.15.6 it seems

guybedford:
Yes it's not yet released yet! Release coming soon.

PR #821, opened 2015-06-09
support overrides within dependencies

jiagang:


guybedford:
Thanks! I think this can be done more simply though by just adding a check for `config.pjson.overrides[resolution.exactName]` at https://github.com/jspm/jspm-cli/blob/master/lib/install.js#L304, which then sets `options.override` if not already set?

lib/package.js
```diff
@@ -473,7 +475,7 @@ exports.download = function(pkg, options, installDeps) {
         if (endpoint.getOverride) {
           if (!override && config.pjson.overrides[pkg.exactName])
             ui.log('warn', 'Using local override for `' + pkg.exactName + '`');
-          return endpoint.getOverride(pkg.registry, pkg.package, pkg.version, override || config.pjson.overrides[pkg.exactName]);
```
guybedford: Surely this line already does everything we want it to do though? How does the new `overrides` object provide more information?

jiagang:
Thank you for your attention. 
The new overrides object just is overrides within dependencies module, and be able to pass into sub dependencies module, it has lower priority than options.override and overrides in root package. What more information is needed?

guybedford:
Right so then is it specifically designed to be used via the API:

```javascript
jspm.install('jquery', { overrides: {} });
```

If so, is there a reason you aren't using the package.json here?

jiagang:
I never thought of it this change to affect install api, I just want to install A module depend B module which with a overrides in B‘s package.json , is there a better way?

guybedford:
@jiagang can you show an example of how you would use this then?

jiagang:
Yes, such as custom module A depend on github:dcodeIO/ProtoBuf.js, because ProtoBuf.js depend on other module, so I define overrides in A's package.json like this
```javascript
"dependencies": {
  "protobufjs": "github:dcodeIO/ProtoBuf.js@4.0.0-b1"
}
"overrides": {
  "github:dcodeIO/ProtoBuf.js@4.0.0-b1": {
    "main": "dist/ProtoBuf",
    "dependencies": {
      "ByteBuffer": "github:dcodeIO/ByteBuffer.js"
    }
  },
  "github:dcodeIO/ByteBuffer.js@3.5.4": {
    "main": "dist/ByteBufferAB",
    "format": "cjs",
    "dependencies": {
      "long": "github:dcodeIO/Long.js"
    }
  },
  "github:dcodeIO/Long.js@2.2.3": {
    "main": "dist/Long"
  }
}
```
Now I install module A in other module B use this
```
  jspm install github:jiagang/A
```
I want to install correctly A with it's all dependencies, but now can't use the overrides in A's package.json.

guybedford:
Right thanks I see, so you want to include the override in the package.json of the published package and have it used?

This is not supported by design actually. The point is that overrides destroy portability and need to be universal. If everyone had custom overrides like this, then dependency conflicts would happen all the time. That is why there is a registry to send a PR with overrides at https://github.com/jspm/registry/ for this purpose.

jiagang:
Thank you for your patience, I get it.

PR #869, opened 2015-06-24
New jspm api updates

guybedford:
It would be nice to get this API update done and tested to go along with the 0.16 release.


guybedford:
(just the docs for the API currently)

Issue #882, opened 2015-06-26
registry overrides don't work on upgrade


lookfirst:
This is my override block:

```
    "overrides": {
      "github:siddii/angular-timer@1.3.3": {
        "main": "dist/angular-timer",
        "registry": "jspm",
        "format": "global",
        "dependencies": {
          "angular": "*",
          "moment": "*",
          "humanize-duration": "github:EvanHahn/HumanizeDuration.js@^3.0.0"
        },
        "shim": {
          "dist/angular-timer": {
            "deps": [
              "angular",
              "humanize-duration",
              "moment"
            ]
          }
        }
      },
      "github:EvanHahn/HumanizeDuration.js@3.0.0": {
        "main": "humanize-duration",
        "format": "global"
      }
    }
```

Of course HumanizeDuration releases a new version... 3.1.0. Now my runtime breaks the next time I do a `jspm install`. Beyond waiting for my PR's to get accepted to the registry or hard locking the version of HumanizeDuration, how do I make things more solid?

guybedford:
Thanks, we could do what we do in the jspm registry and have overrides apply to all versions semver compatible with the override version.

OrKoN:
@guybedford could you point out the relevant places in the code of jspm registry and jspm-cli? I would like to contribute to this.

guybedford:
@OrKoN that would be really great if you are interested in that. It would just be at https://github.com/jspm/jspm-cli/blob/master/lib/package.js#L474.

It would just need to loop through `config.pjson.overrides` to match the `package.name` instead of exact name I think, and then ensure that the version is semver compatible with the currently installing version (package.version).

guybedford:
Released in jspm@0.16.0-beta.6.

PR #891, opened 2015-06-28
Apply overrides to semver compatible versions

OrKoN:
See https://github.com/jspm/jspm-cli/issues/882. Does this require a unit test? I have not found an existing test.


guybedford:
Thanks so much! If you're able to make a test file following the pattern in https://github.com/jspm/jspm-cli/blob/master/test/api.js#L14 for this that would be very welcome to improve coverage here. Otherwise will do manual testing as soon as I can.

lib/package.js
```diff
@@ -471,9 +471,21 @@ exports.download = function(pkg, options, installDeps) {
 
         // get the override
         if (endpoint.getOverride) {
-          if (!override && config.pjson.overrides[pkg.exactName])
-            ui.log('warn', 'Using local override for `' + pkg.exactName + '`');
-          return endpoint.getOverride(pkg.registry, pkg.package, pkg.version, override || config.pjson.overrides[pkg.exactName]);
+          if (!override) {
+            for (var overrideName in config.pjson.overrides) {
+              if (overrideName.indexOf(pkg.name) !== -1) {
```
guybedford: This should be `startsWith` I think?

lib/package.js
```diff
@@ -471,9 +471,21 @@ exports.download = function(pkg, options, installDeps) {
...
+                var overrideVersion = overrideName.split('@')[1];
```
guybedford: A straightforward split is not enough because `@` is an allowed character in package names. So we need to do a `lastIndexOf('@')` substr thing here.

guybedford: Or `overrideName.split('@').pop()`.

lib/package.js
```diff
@@ -471,9 +471,21 @@ exports.download = function(pkg, options, installDeps) {
...
+
+                // use an override if it is compatible per semver
+                if (semver.match(overrideVersion, pkg.version)) {
```
guybedford: This should be `semver.match('^' + overrideVersion)`. The idea is to make the override range semver-based just like the registry, not to allow arbitrary version ranges for overrides. Alternatively do you think we should just do the version ranges and update the registry to be based on version ranges for overrides as well?

OrKoN: I was thinking mostly about the local overrides. For me it would be convenient if I could use any semver pattern in the local override (I am not sure about a use case). I tested on this override:

```
"overrides": {
      "github:components/ember@^1.13.2": {
        "main": "ember.debug",
        "files": [
          "ember.debug.js",
          "ember.prod.js"
        ],
        "dependencies": {
          "jquery": "github:components/jquery@^2.0.3"
        }
      }
    }
```

OrKoN:
I have updated the code according to your comments. Will try to add tests tomorrow.

OrKoN:
I can't image how to test the current version of the code. Added something like that:

```
  test('Override', function(done) {
    api.dlLoader()
      .then(function() {
        return api.install('ember', '1.13.2');
      })
      .then(function() {
        return api.import('ember').then(function(ember) {
          console.log(ember);
          done();
        });
      })
      .catch(done);
  });
```

But how do I check if it used an override?

guybedford:
@OrKoN I guess you could have a case where the override changes the main to a different file, and you check the module matches that alternative main or something like that?

OrKoN:
@guybedford added a test. Please review

lib/package.js
```diff
@@ -471,9 +471,20 @@ exports.download = function(pkg, options, installDeps) {
 
         // get the override
         if (endpoint.getOverride) {
-          if (!override && config.pjson.overrides[pkg.exactName])
-            ui.log('warn', 'Using local override for `' + pkg.exactName + '`');
-          return endpoint.getOverride(pkg.registry, pkg.package, pkg.version, override || config.pjson.overrides[pkg.exactName]);
+          if (!override) {
+            for (var overrideName in config.pjson.overrides) {
+              if (overrideName.startsWith(pkg.name)) {
```
guybedford: A package name might be contained within another package name - eg `github:components/jquery` and `github:components/jq` say. So we should check that `overrideName.startsWith(pkg.name + '@')`.

lib/package.js
```diff
@@ -471,9 +471,20 @@ exports.download = function(pkg, options, installDeps) {
...
+                var overrideVersion = '^' + overrideName.split('@').pop();
```
guybedford: Thanks for updating this.

test/api.js
```diff
@@ -24,4 +25,20 @@ suite('API Calls', function() {
...
   });
+
+  test('Overrides', function(done) {
+    api.dlLoader()
+      .then(function() {
+        return api.install('ember', '1.13.2');
+      })
+      .then(function() {
+        return api.normalize('ember').then(function(normalized) {
+          var content = fs.readFileSync(normalized.replace('file://', ''), 'utf-8');
```
guybedford: In Windows, we need to remove three `///` here. Feel free to define a `fromFileURL` method next to https://github.com/jspm/jspm-cli/blob/master/lib/common.js#L24.

guybedford:
This is looking really great, thanks so much for your work on this @OrKoN.

OrKoN:
Thanks for the review. Updated

lib/common.js
```diff
@@ -25,6 +25,14 @@ exports.toFileURL = function toFileURL(path) {
   return 'file://' + (process.platform.match(/^win/) ? '/' : '') + path.replace(/\\/g, '/');
 };
 
+exports.fromFileURL = function fromFileURL(path) {
+  if (process.platform.match(/^win/)) {
+    return path.replace('file:///', '');
+  } else {
+    return path.replace('file://', '');
+  }
```
guybedford: `return path.substr(process.platform.match(/^win/) ? 8 : 7).replace(path.sep, '/');`

guybedford: Or `path.replace(/^file:\/\/\//, '')` and then the windows conditional.

OrKoN:
Updated fromFileURL. Tests are fine locally but fail on travis. Is something wrong on travis?

guybedford:
Great, thanks for your work on this!

PR #898, opened 2015-07-01
Add cli docs for 'jspm resolve --only'

kenoir:
As documented here: https://github.com/jspm/jspm-cli/wiki/Installing-Packages


theefer:
Helpful!

@guybedford should the CLI also hint at this command when listing forks after an install?

guybedford:
Sure, thanks!

Issue #853, opened 2015-06-17
Conflicting behavior for creating bundle with jspm and loading with systemjs


ahmedshuhel:
The conflicting behavior of `systemjs` and `jspm` bellow are causing problems:

- When jspm creates bundle with `--inject` it creates relative to `baseURL` and does not consider the `paths`.
- When systemjs loads an injected bundle, it loads relative to `paths`(if set) not baseURL. 

Consider the paths for [skeleton-navigation](http://github.com/aurelia/skeleton-navigation)` for example:
```javascript
  "paths": {
    "*": "dist/*.js",
    "github:*": "jspm_packages/github/*.js",
    "npm:*": "jspm_packages/npm/*.js"
  },
```
So when we run `jspm bundle app dist/bundle.js -i` it creates a bundle in the `dist` folder but get's injected in the `config.js` like:

```javascript
  "bundles": {
    "dist/bundle": [
      "npm:core-js@0.9.15/modules/$.fw",
      "npm:core-js@0.9.15/modules/$.dom-create",
 ```
> Note that, it did not consider the `paths`.

Now, when the application tries to load the bundle `systemjs` looks for an url like `dist/dist/bundle` as our `*` path is configured to `dist`. Thus fail to load.

On the other hand,
If we create a bundle with `jspm bundle app bundle.js -i`, this will create the bundle file in `baseURL` not in the `dist` folder, and application will look for the bundle with url like `dist/bundle.js` as the `*` path is set to `dist`. Again, the app will fail to load the bundle.

bundle when injected is treated as a regular module as long as path resolution of `systemjs` is concern. So, it makes more sense if `jspm` considers `paths` along side `baseURL` when `inject` is set to true. 

I have resolved this issue in another project changing jspm api, I would like to participate and make a PR, if you allow me. Thank you.

**Comment by guybedford** [comment]
Perhaps you need to try `jspm bundle app dist/bundle.js` here.

**Comment by ahmedshuhel** [comment]
Are you suggesting without `--inject`? Without inject it is not a problem because we are including the bundle manually. The problem I am having here is `with inject`. 

Considering the above `paths`, if we run `jspm bundle app dist/app-bundle.js -i` it creates a bundle in the dist folder but get's injected in the config.js like:
```javascript
  "bundles": {
    "dist/app-bundle": []
```
Which should be :
```javascript
  "bundles": {
    "app-bundle": []
```
As long as `SystemJS` is concern.

guybedford:
[...]

PR #908, opened 2015-07-05
fix bundle name problem #853

ahmedshuhel:


ahmedshuhel:
@guybedford please review.

lib/bundle.js
```diff
@@ -87,7 +89,7 @@ exports.bundle = function(moduleExpression, fileName, opts) {
 
     if (opts.inject) {
       // Add the bundle to config if the inject flag was given.
-      var bundleName = extractBundleName(fileName);
+      var bundleName = extractBundleName(fileName, systemBuilder);
```
guybedford: Surely we can just make this `systemBuilder.getCanonicalName(fileName)`?

guybedford:
Looks great, please let me know if my comment works out the same.

lib/bundle.js
ahmedshuhel: I believe, the func `systemBuilder.getCannonicalName(...)` needs a proper `file URL` so, if we inline this we probably have to do something like:
``` javascript
 var bundleName = builder.getCanonicalName( toFileURL(path.resolve(config.pjson.baseURL, fileName).replace(/\\/g, '/'))); 
``
This is mostly what `extractBundleName` is doing now.

Please let me know, if you prefer the above approach. I will make the change.

File: lib/bundle.js [hunk 2]
```diff
@@ -173,8 +175,9 @@ function logTree(tree) {
   ui.log('info', '');
 }
 
-function extractBundleName(fileName) {
-  return path.relative(config.pjson.baseURL, fileName.replace(/\.js$/, '')).replace(/\\/g, '/');
+function extractBundleName(fileName, builder) {
+  var file = path.resolve(config.pjson.baseURL, fileName).replace(/\\/g, '/');
```
guybedford: The replacement of `\` with `/` is handled by toFileURL.

lib/bundle.js
guybedford: Oh right, sorry of course. That sounds fine then. I generally like inlining when something isn't used elsewhere as I find it more readable, but happy either way on this.

lib/bundle.js
ahmedshuhel: Actually, you are right here. As `toFileURL` handles `\` replacement, there's nothing much left to `extractBundleName`. Besides, `exaractBundleName` requires `builder` as a parameter, which is somewhat ugly. So, inlining makes much more sense.

Making the changes.
`      var bundleName = systemBuilder.getCanonicalName(toFileURL(path.resolve(config.pjson.baseURL, fileName)));
`

ahmedshuhel:
Inlined `extractBundleName`. Please review.

One test is failing on Travis. I believe, nothing related to bundle/extractBundleName. 

It's trying to login to github but failing:
https://travis-ci.org/jspm/jspm-cli/jobs/70277573

```shell
․․Would you like to set up your GitHub credentials?:Yes
Enter your GitHub username:undefined
Enter your GitHub password or access token:null
Would you like to test these credentials?:Yes
Would you like to try new credentials?:Yes
Enter your GitHub username:undefined
Enter your GitHub password or access token:null
Would you like to test these credentials?:Yes
Would you like to try new credentials?:Yes
```

guybedford:
Excellent, thanks.

ahmedshuhel:
Thank you for your help and support.

PR #922, opened 2015-07-10
Add ability to inject Github auth token from env

kudos:
This makes the Travis/Circle/Heroku flow simpler by only needing to set an environment variable instead of having to add a command to your build process.

Happy to discuss the implementation here and to add docs if when we're satisfied with it.

### DIFF
--- a/lib/global-config.js
+++ b/lib/global-config.js
@@ -115,6 +115,11 @@ if (exports.config.github) {
   dprepend(exports.config.registries.github, exports.config.github);
   delete exports.config.github;
 }
+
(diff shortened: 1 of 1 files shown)


guybedford:
@kudos thanks! Ideally this should be implemented in the GitHub registry itself at https://github.com/jspm/github/blob/master/github.js#L85. Would it be possible to get a PR there?

guybedford:
Also a doc PR would be appreciated here as well when it lands.

kudos:
@guybedford that makes way more sense, still getting familiar with the code :)

PR #941, opened 2015-07-15
log error reason

brillout:
I had an access right problem and logging the error would have helped


guybedford:
Thanks! I merged this manually with an adjustment in https://github.com/jspm/jspm-cli/commit/00db0b464a2685ce4ffeefbdd4cb9f78e7a203c2.

PR #957, opened 2015-07-23
Docs/GitHub rate limits

kudos:
Adds docs for Github auth and rate limiting and removes the Travis CI configuration section that it supersedes. I moved the `JSPM_GITHUB_AUTH_TOKEN` in here too since it makes more sense that way.

Also title-cased some of the headings and my editor trimmed trailing whitespace of its own accord.

### DIFF
--- a/docs/registries.md
+++ b/docs/registries.md
@@ -124,26 +124,6 @@ For npm, you may wish to automate the loading of config from npmrc. This can be 
 
 Which will just respond with defaults to all questions asked during registry install.
 
-#### Travis CI
-
-To configure registries through TravisCI, use the [Travis CLI tool](https://github.com/travis-ci/travis.rb#installation) to encrypt the **[JSPM_GITHUB_AUTH_TOKEN](#auto-configuring-registries)** from the `jspm registry export`.
-
-```
-travis encrypt 'JSPM_GITHUB_AUTH_TOKEN=[JSPM_GITHUB_AUTH_TOKEN]'
-```
-
-Then include it in Travis.yml:
-
-```yml
-env:
-  global:
-  - secure: [ENCRYPTED_STRING]
-
-before_install:
-- npm install -g jspm
-- jspm config registries.github.auth $JSPM_GITHUB_AUTH_TOKEN
-```
-
 ### Creating a Private jspm Registry
 
 You may wish to run your own version of the jspm registry instead of using the publicly maintained default. Running your own registry is particularly useful if you want to create short names to private packages and test lots of overrides.
--- a/docs/registries.md
+++ b/docs/registries.md
@@ -26,12 +26,6 @@ Would you like to test these credentials? [yes]:
 
 This will enable private repo installs.
 
-#### Github Rate Limiting
-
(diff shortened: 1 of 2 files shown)


kudos:
@guybedford as promised :)

docs/registries.md
```diff
@@ -16,16 +16,22 @@ To support private GitHub, simply authenticate with your private GitHub account:
...
 Enter your GitHub username: username
-Enter your GitHub password or access token: 
-Would you like to test these credentials? [yes]: 
+Enter your GitHub password or access token:
+Would you like to test these credentials? [yes]:
 ```
 
 This will enable private repo installs.
 
+#### Github Rate Limiting
```
guybedford: Perhaps `Continuous Integration - Authentication Tokens`

docs/registries.md
```diff
@@ -16,16 +16,22 @@ To support private GitHub, simply authenticate with your private GitHub account:
...
+
+Github rate limits by IP unless an auth token is provided. On hosted services such as Circle, Travis or Heroku where you share IPs with other users also using Github you will likely hit these limits. Setting a `JSPM_GITHUB_AUTH_TOKEN` variable in your environment is enough to tell jspm to use it. Most services have the ability to do this from their web panels.
```
guybedford: It is worth also noting the `jspm config registries.github.auth AUTH_TOKEN` command.

docs/registries.md
```diff
@@ -119,27 +124,7 @@ For npm, you may wish to automate the loading of config from npmrc. This can be
...
-
-```yml
-env:
-  global:
-  - secure: [ENCRYPTED_STRING]
-
-before_install:
-- npm install -g jspm
-- jspm config registries.github.auth $JSPM_GITHUB_AUTH_TOKEN
-```
```
guybedford: I do still like the idea of including direct instructions for users, although perhaps you're right that this is overkill.

guybedford:
I was actually just wondering about it when you sent the PR! Thanks so much for following up, it's really appreciated especially since it is not something that is ever assumed or expected in open source.

docs/registries.md
kudos: I think given that it's possible to just add environment variables via the web panel now, that is the preferred method. In the scenario most people would encounter, they'll first hit the limits on a remote build, and all they'll need to do is add the env var and not check in any changes to their repo.

docs/registries.md
kudos: Yeah, I think it makes sense to say what happens internally when the env variable is encountered.

docs/registries.md
kudos: I had used `Continuous Integration` first, but this also applies to services like Heroku as well as people just randomly behind proxies and firewalls.

kudos:
@guybedford I'm a big fan of jspm and systemjs, happy to help where I can :)

docs/registries.md
guybedford: Right, I just don't want people to confuse this for normal authentication which should be communicated to users first via `jspm registry config github`. That's also why this should be lower down with a clear title that shows it is specific to automated workflows.

docs/registries.md
guybedford: Sure, that makes sense!

kudos:
@guybedford I reworded and moved the rate limiting out of registries.md entirely and into installing-packages.md. I also added some SEO juice by including the error message jspm outputs on the doc page.

guybedford:
Thanks, I manually merged this in https://github.com/jspm/jspm-cli/commit/6c8752c8d09c8668f90f0b03b39da146a9c683ad, bringing it up further to the top as well.

PR #965, opened 2015-07-28
Set meta information at the end so that entries get normalized properly

matoilic:
Currently, when the config file is written meta information is output before map. If you set meta information for installed libraries, e.g. adding jQuery as dependency for Angular, the meta entry for would not be normalized properly and thus would not have any effect. This change writes the meta config at the end of the SystemJS configuration.

### DIFF
--- a/lib/common.js
+++ b/lib/common.js
@@ -143,9 +143,18 @@ exports.getRedirectContents = function(format, main) {
     throw 'Unknown module format ' + format + '.';
 };
 
-exports.stringify = function (subject) {
+exports.stringify = function (subject, contentOnly) {
   // replace the LF used by `JSON.stringify` with the preferred new line
-  return JSON.stringify(subject, null, 2).replace(/\n/g, EOL);
+  var str = JSON.stringify(subject, null, 2).replace(/\n/g, EOL);
+
+  if(contentOnly) {
+    //remove curly braces around object properties
+    str = str.split(EOL)
+      .slice(1, -1)
+      .join(EOL)
+  }
+
+  return str;
 };
 
 /* Recursively remove directory, all those above it, if they are empty.
--- a/lib/config.js
+++ b/lib/config.js
@@ -29,6 +29,8 @@ var EOL = require('os').EOL;
 // default newline to the appropriate value for the system
 exports.newLine = EOL;
 
+exports.tab = '  ';
+
 // given a package.json file and an override
 // calculate the package.json file jspm will see
 exports.derivePackageConfig = function(pjson, override) {
--- a/lib/config/loader.js
+++ b/lib/config/loader.js
@@ -352,14 +352,14 @@ Config.prototype.write = function() {
   cfg.map = map;
   cfg.versions = versions;
 
-  if (hasProperties(meta))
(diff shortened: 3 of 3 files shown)


Munter:
Is this the reason the map shorthands could not be used to target jspm installed library labels?

guybedford:
SystemJS's configuration function will always ensure that map is set before meta (https://github.com/systemjs/systemjs/blob/0.18.4/lib/core.js#L145). Are you using a version before SystemJS 0.18.4?

matoilic:
No, I'm using the latest beta of JSPM with SystemJS 0.18.4.

guybedford:
Ahh, because these are separate configuration calls we do need this PR.

I'm wondering if we should just make the config a single call then anyway while we're at it? (because its nice having user properties at the top)

We could even remove first-level quoting of properties as well, something like - 

```javascript
System.config({
  baseURL: ".",
  defaultJSExtensions: true,
  transpiler: "babel",
  babelOptions: {
    "blacklist": [
      "es6.arrowFunctions"
    ],
    "optional": [
      "runtime"
    ]
  },

  paths: {
    "github:*": "jspm_packages/github/*",
    "npm:*": "jspm_packages/npm/*"
  },

  packages: {
    "app": {
      "defaultExtension": "js",
      "meta": {
        "*.json": {
          "loader": "json"
        }
      }
    },
    "asdf": "asdf"
  },

  meta: {
    "*.html": {
      "loader": "text"
    }
  },

  /*
   * Map Config
   * This represents the exact version application lock
   */ 
  map: {
    "key": "value" // ...
  }
});
```

guybedford:
Could be nice to introduce more comments to explain the config to users.

matoilic:
It would make sense to merge the whole config into one call to prevent such issues in the future. And a few comments wouldn't hurt :). Should I integrate it into this PR?

guybedford:
@matoilic if you are interested in updating the PR here along these lines that would be very welcome!

matoilic:
I've merged the configuration into one config call and removed quotes around property names where possible. Also, on a second thought, the comments are not as practical as thought. Neither can additional comments be added so that they're preserved nor can generated comments permanently be removed. Each install call would reset all changes to comments.

guybedford:
Thanks so much, I've manually merged this.

peteruithoven:
Is it normal that is puts the `packages` and `meta` above the `map`, but bellow the `paths` configuration? 
This makes it that JSPM reads the `map` before it reads the `packages` and `meta`, that's quite confusing?

guybedford:
Order doesn't matter now - SystemJS will apply configs in the right order it needs regardless of the order in the configuration object.

peteruithoven:
But isn't that what they call "magic", in that users might not know this and become unsure about what will happen? Since you generate the output (and regenerate it with a JSPM install), doesn't it make sense to sort it?

guybedford:
We include user configs at the top, map configs at the bottom because it's more bulk. JS object property order should never be assumed, as its not specified. That's one of the first thing you learn when using objects.

peteruithoven:
Very valid points.

PR #1006, opened 2015-08-10
correct baseURL default

whitneyit:
When a `config.js` file is created via `jspm init`, the following prompt
is show:

> Enter client baseURL (public folder URL)[./]

However when just hitting enter, the value of `"/"` is entered. This PR
ensures that the default value of `"./"` is used.

### DIFF
--- a/lib/config/loader.js
+++ b/lib/config/loader.js
@@ -184,7 +184,7 @@ Config.prototype.read = function(prompts, sync) {
   if (sync)
     throw 'Configuration file has not been initialized. Run jspm init first.';
 
-  return ui.input('Enter client baseURL (public folder URL)', self.baseURL || '/')
+  return ui.input('Enter client baseURL (public folder URL)', self.baseURL || './')
   .then(function(baseURL) {
     self.baseURL = baseURL;
 


whitneyit:
These tests pass locally. There seems to be some problem with some GitHub integration.

> Enter your GitHub username:undefined
> Enter your GitHub password or access token:null
> Would you like to test these credentials?:Yes
> Would you like to try new credentials?:Yes
> Enter your GitHub username:undefined
> Enter your GitHub password or access token:null
> Would you like to test these credentials?:Yes
> Would you like to try new credentials?:Yes
> Enter your GitHub username:undefined
> Enter your GitHub password or access token:null
> Would you like to test these credentials?:Yes
> Would you like to try new credentials?:Yes

There are hundreds of lines like that.

guybedford:
The baseURL set to `/` is actually encouraged. If serving jspm from a subfolder, it is best to set the baseURL to `/subfolder` so the app works across all URLs instead of just URLs at the right depth level.

Thanks - for the tests I fixed this with an auth token.

whitneyit:
Ok well then can we update what is displayed inside the `[]`. Because It looks like it should be setting the value to the current directory, but `/` is set instead.

guybedford:
The server baseURL is set to `./`. It is the client baseURL that is set to `/`. The prompt for the client baseURL certainly says `/` as far as I can tell.

whitneyit:
You're right. I did get therm mixed up.

If it is best to serve from a folder, do you want me to send a PR to change
it to `/public ` or something similar?

guybedford:
Thanks, happy with the default as it is - for client apps serving from the root can be fine. It's only when mixing server logic with that that one needs to worry.

PR #1008, opened 2015-08-10
Docs: Add more github access token info

peteruithoven:


guybedford:
Thanks, I manually updated the inline prompt here to be correct in https://github.com/jspm/jspm-cli/commit/6c8752c8d09c8668f90f0b03b39da146a9c683ad.

Issue #947, opened 2015-07-17
Lock transpiler version


OliverJAsh:
Is this possible? At the moment it comes down with `jspm dl-loader`. We want the option of upgrading manually. (Developers on the team are getting annoyed that they keep finding changes to the config file.)

desbo:
+1

We'd like to require a specific babel version in `package.json` and update it explicitly. Using `dl-loader` currently overwrites whatever version you have set to `^5.6.4`, so patch updates are automatically installed and written to the config.

guybedford:
Going with a full lock from now on, apologies for the current patch regressions.

https://github.com/jspm/jspm-cli/commit/36abe969b2e0936680bd114f4be251e6b876b6a3

guybedford:
Released in jspm@0.16.0-beta.6.

OliverJAsh:
Thanks for looking into this. This doesn't lock the transpiler, it just locks SystemJS?

guybedford:
Right, yes this was the wrong issue to mark above. In terms of locking the transpiler, the minimum version of the transpiler does need to be enforced by jspm, but then within those bounds, the exact version should be the version the user wants to install and we shouldn't force upgrades unnecessarily.

I'm thinking something along the lines of `jspm dl-loader` to retain whats in the config.js version lock (as long as it is in range) and `jspm dl-loader --latest` to do the full latest upgrade of the transpiler.

guybedford:
Contributions welcome!

desbo:
@guybedford could you take a look at https://github.com/jspm/jspm-cli/pull/1029? It'd solve this problem for us.

guybedford:
@desbo thanks, set as the default behaviour in https://github.com/jspm/jspm-cli/commit/202a1cd9f178a94d7fc30f31a4e61ae93e059376.

guybedford:
Released in 0.16.1.

PR #1029, opened 2015-08-18
add no-update flag to dl-loader

desbo:
Allows you to use `dl-loader --no-update` to download the transpiler version specified in `package.json` without writing any changes to that file. 

See #947

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -375,10 +375,10 @@ process.on('uncaughtException', function(err) {
       break;
 
     case 'dl-loader':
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks so much for this, as mentioned I do agree this should actually be the default though - I've added https://github.com/jspm/jspm-cli/commit/202a1cd9f178a94d7fc30f31a4e61ae93e059376.

desbo:
Ah good, thanks for looking into it.

Issue #528, opened 2015-02-18
Non-verbose mode


guybedford:
How about:

```
  jspm install --quiet
```

or something like that for log-levels?

jackfranklin:
@guybedford happy to take a stab at this :+1: Would quiet mute all messages?

guybedford:
@jackfranklin yes, please do! I'm not that familiar with log-levels actually, are there any good examples out there for how to approach the use cases?

jackfranklin:
@guybedford did some reading into this and found various StackOverflow posts including [this one](http://stackoverflow.com/questions/7839565/logging-levels-logback-rule-of-thumb-to-assign-log-levels).

I quite like this set of logs:

- Error: critical logical errors on application, like a database connection timeout. Things that call for a bug-fix in near future
- Warn: not-breaking issues, but stuff to pay attention for. Like a requested page not found
- Info: used in functions/methods first line, to show a procedure that has been called or a step gone ok, like a insert query done
- log: logic information, like a result of an if statement
- debug: variable contents relevant to be watched permanently

Where Error, Warn and Info are shown by default. You could then allow a `--log` CLI argument which could be set to `error`, `warn` and so on. When set, jspm would only log anything for that log level and up (so `--log warn` shows warnings and errors). Then, `--log debug` effectively becomes verbose mode, and `--log error` becomes non-verbose.

Does something like that sound vaguely sensible?

guybedford:
Sorry for the delay on this, have just been thinking about it and this proposal sounds ideal.

Just let me know if you have further implementation questions at all.

dvc94ch:
Is a solution this complicated really needed?

I'm thinking of checking for the debug flag and if it's there don't throw a new error and let the old one bubble up.
Or do you think it would be better solved by adding a `debug` function to ui and then add a filter to ui.log?
one could then ui.log('debug', error) before throwing a new one

dvc94ch:
If taking the second route, what would be the default log level? Everything except debug?

sergei-maertens:
another go at it might be using the -v, -vv, -vvv and -vvvv verbosity flags. That's how ssh does it if I'm not mistaken.

Anyway, big +1 from me for being able to surpress anything but errors!

watilde:
Has this issue already resplved..?

jackfranklin:
@watilde, with PR #1082 it was, yes. The next release of jspm will support a `--log` option :tada:

watilde:
[...]

PR #1082, opened 2015-09-01
Include the ability to customise log levels

jackfranklin:
This PR adds four new log levels, customised with the `--log` flag.

- `--log debug` (default) everything is logged
- `--log info` info, warnings and errors are logged ('ok' is excluded)
- `--log warn` warnings and errors are logged
- `--log err` only errors are logged

@guybedford I finally found some time to look at #528 and this is my first stab. What do you think?

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -88,8 +88,8 @@ process.on('uncaughtException', function(err) {
       + 'jspm dl-loader [babel|traceur|typescript]\n'
       + '\n'
       + 'jspm resolve --only registry:package@version\n'
-      + '  resolve --only npm:jquery@2.1.1  Resolve all versions of a package to the given version\n'    
-      + '\n'  
+      + '  resolve --only npm:jquery@2.1.1  Resolve all versions of a package to the given version\n'
+      + '\n'
       + 'jspm setmode <mode>\n'
       + '  setmode local                    Switch to locally downloaded libraries\n'
       + '  setmode remote                   Switch to CDN external package sources\n'
@@ -205,6 +205,12 @@ process.on('uncaughtException', function(err) {
 
   var args = process.argv.splice(2),
       options;
+
+  var logArgIndex = args.indexOf('--log');
+  if (logArgIndex > -1) {
+    ui.setLogLevel(args[logArgIndex + 1]);
+    args.splice(logArgIndex, 2);
+  }
 
   switch(args[0]) {
     case 'run':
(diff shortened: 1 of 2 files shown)


cli.js
```diff
@@ -206,6 +208,11 @@ process.on('uncaughtException', function(err) {
   var args = process.argv.splice(2),
       options;
 
+  options = readOptions(args);
```
guybedford: Perhaps just use `readOptions(args, [], ['log'])` here to read the log level.

cli.js
```diff
@@ -142,6 +142,8 @@ process.on('uncaughtException', function(err) {
 
     flags = flags || [];
     optFlags = optFlags || [];
+    // we always take a log option
+    optFlags.push('log');
```
guybedford: This is crossing concerns, see the suggestion below.

guybedford:
This is amazing! Thanks so much Jack for taking the time to put this together. Will definitely merge this in.

I'm wondering if we should make the default log level `info` instead of `debug` and include `ok` messages in that as well? Perhaps an `ok` log level could then just include `warn`, `err`, `ok` for a reduced log.

cli.js
jackfranklin: We need to have this here, or add it to every `readOptions` call, else you get a `Warn: unknown option log` at the end of every call to a jspm command.

cli.js
jackfranklin: :+1:

jackfranklin:
@guybedford so if I'm understanding correctly:

- `info` = default, has everything
- `ok` = everything __except info__
- `warn` = as before, warn and err
- `err` = just error

cli.js
guybedford: Ah right of course. I know it's not ideal, but perhaps we just need to strip out the `--log` argument manually instead of running `readOptions` twice, something like:

```javascript
  var logArg = args.indexOf('--log');
  if (logArg > -1) {
    ui.setLogLevel(args[logArg + 1]);
    args.splice(logArg, 2);
  }
```

guybedford: I guess global arguments will always be a thing though, so the architecturally correct solution would probably be to wrap `readOptions` with some default global handlings within a new function closure.

guybedford:
:+1:

cli.js
jackfranklin: Went with your code above for now; I think it probably requires a bit of a rework the next time we add a global option, but for now it seems overkill.

jackfranklin:
@guybedford new version pushed up with al the changes! Is there anything else that needs doing?  A line in the help?

guybedford:
:+1: This is great, just added https://github.com/jspm/jspm-cli/commit/fa046c3e2ee6de5ac88298b641a68bf13ae447f7. Sounds sensible to abstract the global layer properly only when needed. Thanks for your efforts!

PR #1086, opened 2015-09-02
Update default log level from debug to info

jackfranklin:
@guybedford I missed this in the original PR :cry:

### DIFF
--- a/lib/ui.js
+++ b/lib/ui.js
@@ -22,7 +22,9 @@ var logging = true;
 var logBuffer = '';
 var logged = false;
 var cli = module.exports;
-var logLevel = 'debug';
+var DEFAULT_LOG_LEVEL = 'info';
+
+var logLevel = DEFAULT_LOG_LEVEL;
 
 var logMap = {
   ok: ['ok', 'warn', 'err'],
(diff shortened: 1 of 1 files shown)


lib/ui.js
```diff
@@ -95,7 +96,7 @@ exports.log = function(type, msg) {
     msg = format[type](msg);
 
   if (logging) {
-    if ((logLevel === 'debug') ||
+    if ((logLevel === DEFAULT_LOG_LEVEL) ||
```
jackfranklin: this is so any messages without a `type` get shown only when the log level is at the default

guybedford:
Ah, caught this in https://github.com/jspm/jspm-cli/commit/fa046c3e2ee6de5ac88298b641a68bf13ae447f7, so we should be ok?

jackfranklin:
I think that will mean that any message that doesn't have a type will now never be logged? I think you still need to have that conditional that just logs absolutely anything if the log level is `info`.

guybedford:
@jackfranklin https://github.com/jspm/jspm-cli/commit/bc5ad99c2095e40c03bfae23c33f5f51ac2d91fb? Let me know if I'm missing something here though.

jackfranklin:
Ah yeah, I think that covers it :+1:

guybedford:
Great, thanks!

PR #1087, opened 2015-09-02
do not clean if packages directory is a symlink

stefanr:


lib/install.js
```diff
@@ -947,6 +947,17 @@ function clean() {
...
+function clean() {
+  return new Promise(function(resolve, reject) {
+    fs.lstat(config.pjson.packages, function(err, stats) {
+      if (err) {
+        reject(err);
+      } else {
+        resolve(stats.isSymbolicLink() ? true : _clean());
+      }
+    });
+  });
```
guybedford: We don't need to skip the entire clean, we just need to skip steps that alter `jspm_packages`, which is 4 and 5 above.

guybedford:
Thanks for having a go at this one!

stefanr:
Only skipping cleaning steps 4 and 5 for symlinked jspm_packages ...

guybedford:
Amazing, thanks so much for this!

PR #1099, opened 2015-09-06
0.17 wip

guybedford:
Alpha Release Checklist
- [x] New dynamic package configuration system
- [x] New linking workflow
- [x] Endpoint upgrades to configuration system
- [x] Persistent overrides for override reproducibility with no invisible overrides
- [x] Peer dependency support
- [x] Transpilers as plugins
- [x] Config file management to fully retain style and properties
- [x] GitHub registry endpoint to no longer use GitHub releases (NB requires checking things like Bootstrap and d3 overrides)
- [x] Full support for server-side npm installs
- [x] npm registry endpoint to support any module format
- [x] Local project as a package itself with init prompt handling
- [x] Complete Babel 6 plugin (https://github.com/systemjs/plugin-babel)
- [x] Complete Traceur plugin (https://github.com/systemjs/plugin-traceur)
- [x] Automatic Rollup optimizations for jspm build (https://github.com/systemjs/builder/pull/442)
- [x] Upgrade path testing
- [x] Comprehensive build testing


djindjic:
sorry @guybedford, I didn't catch what does it means if `GitHub registry endpoint to no longer use GitHub releases`? master branch will be pulled? probably I'm wrong

guybedford:
@djindjic we no longer download the separate release asset file uploads, but always use the git tree tagged or branch releases.

nmschulte:
These changes affect linking: linked packages without a `"registry"` property declared fail to link; with 0.16 they would link without issue.  Declaring `"jspm"` didn't actually work in my case: it turns out, this particular package actually does use npm for its `dependencies` registry/namespace, but the dependencies are unimportant for the module(s) I'm using from the package.  It's a fairly edge-case, but worth noting I hope.

guybedford:
@nmschulte I think this is actually quite an important error, because the idea is to replicate the way the package would behave if installed from a registry. If you published that package to github, then it would try to install those dependencies you don't want and fail, making a difference between the linking behaviour and package behaviour, which is exactly what we're trying to avoid.

edrex:
I noticed that `setmode remote` is removed in f52bdf5842c13e91c9809761656308502137928b. Is loading directly from the CDN not supported anymore in 0.17?

djindjic:
@guybedford do you have any prediction when will 0.17 become rc or main release? I'm following this pr to get this info :)

guybedford:
@djindjic the priority is SystemJS documentation and then the jspm 0.17 documentation and then the release can happen. I estimate a week of work for each, and with my current schedule that should take about 2 months to find the time.

theodorton:
@guybedford Anything that may be contributed with regards to the docs?

JulianIsrael:
Hi, should 0.17 work in linux?, I have tried and installing packages is not working jspm install npm:react->
Error: unexpected end of file at Zlib._handle.onerror (zlib.js"370:17) and so on similar errors in other libraries.

guybedford:
jspm 0.17 is officially deprecated.

Issue #1109, opened 2015-09-09
Document bundle dependencies functionality


peteruithoven:
According to the following posts it should be possible to bundle dependencies:
https://github.com/jspm/jspm-cli/issues/482#issuecomment-86668706
https://github.com/jspm/jspm-cli/issues/872#issuecomment-114946729
This is of course totally awesome, but I had to find out trough the issues queues and it seems to use an undocumented syntax. 

Bundling the dependencies should be possible with the command: 

``` bash
jspm bundle app/**/* - [app/**/*]
```
- How does `app/**/*` describe all the dependencies in `jspm_packages` but not one's custom code? 
- `app` should be one's main javascript? So when this is located in a _lib_ folder it should be: `jspm bundle lib/app/**/* - [lib/app/**/*]`?
- What's that `- [app/**/*]` portion? 
- In both cases @guybedford mentions this functionality `--inject` isn't used, is there a reason for that? 

Maybe we could add this info to a _Development workflows_ doc page?

peteruithoven:
I'm using JSPM version `0.16.2`, which seems more recent than the latest beta. 

I've created a new jspm enabled project with `jspm init`, in which I installed `lodash`. I created `app.js` which uses lodash. Then I tried bundling my dependencies, in this case only `lodash` with the following command:
```
jspm bundle app/**/* - [app/**/*] bundle.js 
```
But the created `bundle.js ` file is empty. There is no `bundle.js.map` created. 
Only when I bundle the actual app with `jspm bundle app bundle.js` the bundle.js is filled. 

I've published this example: https://github.com/peteruithoven/jspm-bundle-deps

OrKoN:
Do you use bash? I remember someone had an issue in some other shell because additional escaping was needed. Also simply use gulp to bundle dependencies: 

```
gulp.task('js:bundle', function() {
  var pkg = require('./package.json');
  var deps = pkg.jspm.dependencies;
  var modules = Object.keys(deps);
  var moduleList = modules.join(' + ');
  console.log('\n' + moduleList + '\n');
  return jspm.bundle(moduleList, 'main-bundle.js', { mangle: false, sourceMaps: true, inject: true }).then(function() {
    console.log('Bundled');
  });
});
```

peteruithoven:
Thanks @OrKoN, I tried adding some escaping and I added these to my previous comment. 

[...]

PR #1136, opened 2015-09-18
Add development workflow docs

peteruithoven:
See: https://github.com/jspm/jspm-cli/issues/1109#issuecomment-141371788

### DIFF
--- a/docs/development-workflows.md
+++ b/docs/development-workflows.md
@@ -1,0 +1,10 @@
+### Performance
+There are a couple of things that can be done to speed up the development workflow.
+For more see: https://github.com/jspm/jspm-cli/issues/872
+
+#### Creating a dependencies bundle
+To speed up the module loading and possibly transpilation it's possible to prebuild and bundle all the dependencies (packages in jspm_packages).
+This is possible using the `jspm bundle` feature. If you're code lives in a folder called `lib` you can bundle and [inject](https://github.com/jspm/jspm-cli/blob/master/docs/production-workflows.md#loading-a-bundle-automatically-inject) the dependencies with:
+```
+jspm bundle 'lib/**/* - [lib/**/*] --inject'
+```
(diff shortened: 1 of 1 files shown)


guybedford:
:+1: it would be good to cover the different aspects including server, compilation and caching workflows as well.

peteruithoven:
Definitely! 
But as I'm trying to circumvent compilation steps (with gulp etc) in my development workflow I'm probably not the one to write those.

lookfirst:
@peteruithoven How do you deal with .map files for your code are you just writing everything in ES5?

lookfirst:
There is NO way that bundling would work for me... it is WAY too slow with the number of files that I'm loading. Tens of seconds without minification on and about a minute otherwise. I've got my app split into 4-5 different apps that need individual bundling (and growing constantly)... this wouldn't work at all except for small stuff.

I also don't see the point in skipping the gulp step when bundling is effectively the same thing and you get way more features.

peteruithoven:
I write ES6, `jspm bundle` creates a .map for me. I don't see the problem? 
@lookfirst which part is slow for you? Running the code in the browser or building the bundle? How long it takes to build the dependencies bundle doesn't really matter since you only have to do this after installing or removing dependencies right? I don't have to bundle when changing my own code (during development at least).

stoiet:
Hy! Currently I use 2 commands with exeSync in gulp tasks:
```
jspm bundle source/javascript/**/*.js - [source/javascript/**/*.js] client/dist/bundle-dependency.js --inject --minify --no-runtime --skip-source-maps
jspm bundle  [source/javascript/**/*.js] client/dist/bundle-application.js --inject --no-mangle --inline-source-maps
```

I only watching source/javascript/**/*.js files and re-bundle them with the second command, it would be faster with some caching mechanism, and only use the Builder API for it. With a gulp plugin would be nicer, too.

@lookfirst 's solution seems to better for larger projects, but works only with browserSync.

lookfirst:
I can't imagine doing development without browserSync (you really want to hit reload in the browser every time you change something?!) and size doesn't really matter. =)

lookfirst:
@peteruithoven I don't bundle my dependencies separately. No need with my build system. =)

peteruithoven:
Maybe we can merge these docs and have @lookfirst do a pull request with his approach?

guybedford:
Closing for lack of activity. Feel free to reopen.

PR #1149, opened 2015-09-22
update broken links in registries.md

hmps:
Found a couple of broken links in registries.md. Wasn't sure about the one that pointed to community projects, but the site with third-party resources was the best match I could find.


guybedford:
Thanks so much, merged manually.

PR #1152, opened 2015-09-23
Restore ability to have custom/temporary baseURL on build time

Rush:
I was relying on this functionality:
```js
var tmpBuildDir = '.gulp-build`';
var builder = new jspm.Builder({ baseURL: path.resolve(tmpBuildDir) });
```

What I am doing is I am copying files from multiple source directories to one place and on every file before passing it to gulp I am running ng-annotate filter (https://github.com/olov/ng-annotate). This temporary build dir then has to be used as baseURL.

Anyway, this fixes the behavior I need for my project.

### DIFF
--- a/lib/bundle.js
+++ b/lib/bundle.js
@@ -27,7 +27,11 @@ var toFileURL = require('./common').toFileURL;
 // just allows cfg object
 function Builder(_config) {
   config.loadSync();
-  SystemJSBuilder.call(this, toFileURL(config.pjson.baseURL));
+  var customBaseURL;
+  if (typeof _config == 'object')
+    customBaseURL = _config.baseURL;
+
+  SystemJSBuilder.call(this, toFileURL(customBaseURL || config.pjson.baseURL));
 
   var cfg = config.loader.getConfig();
 


guybedford:
You should be able to achieve this same effect by using wildcard paths that later in build. You can then configure this via `builder.config({ paths: { '*': '.tmp/*'  } })` for example.

Rush:
Unfortunately it's not the case, I tried your snippet and I get:
```
Error: Error: ENOENT, open '/code/project/webapp/frontend/.gulp-build/github:johnsoftek/plugin-jade@0.5.1.js'
    at Error (native)
        Error loading file:///code/project/webapp/frontend/.gulp-build/github:johnsoftek/plugin-jade@0.5.1.js
    at /code/project/webapp/frontend/node_modules/systemjs-builder/lib/builder.js:23:9
    at Object.lib$rsvp$events$$default.trigger (/code/project/webapp/frontend/node_modules/systemjs-builder/node_modules/rsvp/dist/rsvp.js:245:13)
    at null._onTimeout (/code/project/webapp/frontend/node_modules/systemjs-builder/node_modules/rsvp/dist/rsvp.js:779:47)
    at Timer.listOnTimeout (timers.js:119:15)
```
So it applies the paths trick also to any jspm_packages coming from repos, which obviously breaks them

Rush:
Also I want to underline that my fix simply tries to restore previous behavior, with which I was able to change `baseURL` when creating the builder.

guybedford:
@Rush ah nice, I was just trying to find this issue - fixed that one in https://github.com/ModuleLoader/es6-module-loader/commit/2e17f6cdb89bd02b19a47375bfc0370d1e1b0dec.

Rush:
@guybedford nice, how can I test? I am pretty lost at how the dependency tree for whole jspm is handled.

guybedford:
@Rush I'll post up a build a little later today, and then `jspm dl-loader --edge` would do it.

guybedford:
@Rush this has been released in jspm 0.16.10.

PR #1165, opened 2015-09-27
Expose the jspm version in the API

jackfranklin:
Couple of times I've written small tools in which I've wanted to just
log the current version of jspm.

@guybedford seemed like a sensible addition to me - what do you think?

### DIFF
--- a/api.js
+++ b/api.js
@@ -56,6 +56,7 @@ API.promptDefaults = function(_useDefaults) {
   ui.useDefaults(_useDefaults);
 };
 
+API.version = require('./package.json').version;
 
 /*
  * Loader API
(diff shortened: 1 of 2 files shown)


jackfranklin:
I added a test primarily so I had confidence that I had done it right, but the test seems pretty brittle to me, so if you'd like me to get rid of it I'm happy to!

guybedford:
Looks great thanks Jack! A test is fine, let's just make it upgradable by doing `version.match(/\d+\.\d+\.\d+(-[^\.]+)?/)` or something like that?

jackfranklin:
@guybedford perfect, updated :+1:

guybedford:
:+1: (sorry just missed this one for the release!)

PR #1175, opened 2015-10-02
Adds uglify's dead_code&global_defs support for consistency with SystemJSBuilder.

chenzhenxi:
I can't do conditionally drop code while doing minify bundle via JSPM CLI, which SystemJsBuilder original supported.  

See  

https://github.com/systemjs/builder/blob/master/lib/output.js#L82

at line: 82  

    ast = ast.transform(uglify.Compressor({
        dead_code: true,
        global_defs: globalDefs,
        warnings: false
    }));

So I added this missing part to JSPM.

Adds uglify's dead_code and global definition support to CLI for consistency with SystemJS builder.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -94,8 +94,8 @@ process.on('uncaughtException', function(err) {
       + '  setmode local                    Switch to locally downloaded libraries\n'
       + '  setmode remote                   Switch to CDN external package sources\n'
       + '\n'
-      + 'jspm bundle moduleA + module/b [outfile] [--minify] [--no-mangle] [--inject] [--skip-source-maps] [--source-map-contents]\n'
-      + 'jspm bundle-sfx app/main [outfile] [--format <amd|cjs|global>] \n'
+      + 'jspm bundle moduleA + module/b [outfile] [--minify] [--global-defs <json string>] [--no-mangle] [--inject] [--skip-source-maps] [--source-map-contents]\n'
+      + 'jspm bundle-sfx app/main [outfile] [--format <amd|cjs|global>] [--minify] [--global-defs <json string>] \n'
       + 'jspm unbundle                      Remove injected bundle configuration\n'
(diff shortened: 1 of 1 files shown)


cli.js
```diff
@@ -94,8 +94,8 @@ process.on('uncaughtException', function(err) {
       + '  setmode local                    Switch to locally downloaded libraries\n'
       + '  setmode remote                   Switch to CDN external package sources\n'
       + '\n'
-      + 'jspm bundle moduleA + module/b [outfile] [--minify] [--no-mangle] [--inject] [--skip-source-maps] [--source-map-contents]\n'
-      + 'jspm bundle-sfx app/main [outfile] [--format <amd|cjs|global>] \n'
+      + 'jspm bundle moduleA + module/b [outfile] [--minify] [--define <json string>] [--no-mangle] [--inject] [--skip-source-maps] [--source-map-contents]\n'
```
guybedford: Perhaps lets call it also `global-defs`.

lib/build.js
```diff
@@ -579,7 +579,9 @@ exports.compileDir = function(dir, options) {
 
               ast = ast.transform(uglify.Compressor({
                 warnings: false,
-                evaluate: false
+                evaluate: false,
+                dead_code: true,
+                global_defs: options.globalDefs
```
guybedford: This is unrelated and can be left out.

guybedford:
Thanks for the PR. I'm somewhat torn on this though as the build API at https://github.com/jspm/jspm-cli/blob/master/docs/api.md#class-builder is the recommended way of doing more flexible / larger builds over the CLI use case.

chenzhenxi:
Thanks for comments.
I renamed Define to global_defines and removed changes in compile.

I thinking its convenience to have CLI to do more things.

guybedford:
Sure thanks, I can take this but will not be able to document it as a stable API feature.

PR #1239, opened 2015-10-29
Tests fail under windows (CRLF != LF)

ebpa:
Files generated by tests contain Windows-style line endings which fail assert equal tests.  These should either be normalized at write or within the unit tests.

Also changed license format so npm doesn't complain.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,5 +57,5 @@
     "testlibs",
     "test"
   ],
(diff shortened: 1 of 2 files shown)


ebpa:
I guess what npm likes nvm throws a tantrum about?

guybedford:
:+1:

Issue #190, opened 2014-10-17
Local configuration support


guybedford:
It should be possible to have a `.jspmconfig` or similar file in the current project directory, which overrides the global jspm configuration.

This is useful for project-specific credentials / settings.

When running `jspm config`, it would be useful to set the level of setting with `jspm config --local` for setting local config.

**Contribution Steps:**
1. Alter the global config loading code at https://github.com/jspm/jspm-cli/blob/master/lib/global-config.js#L41 to also check for a local `.jspmconfig` file in the current folder (taken to be `process.env.jspmConfigPath || process.cwd()`).
2. Have this local config override the local config
3. Add an option to `jspm config` and `jspm endpoint` (https://github.com/jspm/jspm-cli/blob/master/cli.js#L470, https://github.com/jspm/jspm-cli/blob/master/cli.js#L410) for a `--local` flag to indicate that the [global config save function](https://github.com/jspm/jspm-cli/blob/master/lib/global-config.js#L35) should save the new changes to this local file. Some type of tracking code would be needed to ensure we save just those properties which have been changed.

**Testing:**
- I should be able to do `jspm endpoint create custom jspm-github --local` and create a custom GitHub endpoint that I can see in my local `.jspmconfig` file only and not in `~/.jspm/config`
- I should be able to do `jspm config custom.property value --local` and see my `.jspmconfig` file have this property added, but not `.jspmconfig`.

djindjic:
is there some workaround (for travisci)? something like:

jspm endpoint config github --username=MY_USERNAME --token=MY_TOKEN

guybedford:
@djindjic yes, look at your local `~/.jspm/config` file and you'll see your github auth string.

Then write:

```
  jspm config endpoints.github.auth AUTH_STRING
```

djindjic:
thanks!

andreasgrimm:
(sorry if too obvious) for public repos use the travis cli tool and `travis encrypt 'GH_TOKEN=g5489ugh9...` your github auth token and include it like this in your .travis.yml file:
```yml
env:
  global:
  - secure: "HJJH9889JK......"

before_install:
- npm install -g jspm ...
- jspm config endpoints.github.auth $GH_TOKEN
```

djindjic:
thanks @andreasgrimm, and thanks for sharing your [jspm ng-next-example](https://github.com/ng-next/ng-next-example) sample.
[...]

PR #1265, opened 2015-11-05
Local config

hmps:
This is a proposed fix for issue #190. 

It creates a `--local` flag for `jspm config` and `jspm registry`. If the `--local` flag is provided the configuration will be saved in `./.jspmrc`or `./.config/.jspmrc`, in that order.

I couldn't really wrap my head around if and how to add tests for the changes, so no tests are added. I'd be happy to fix this if there is a standard to follow.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -571,9 +571,11 @@ process.on('uncaughtException', function(err) {
 
     case 'c':
     case 'config':
-      var property = args[1];
-      var value = args.splice(2).join(' ');
-      globalConfig.set(property, value);
+      options = readOptions(args, ['local'])
+      var property = options.args[1];
+      var value = options.args.splice(2).join(' ');
+
+      globalConfig.set(property, value, options.local);
       break;
 
     case 'cc':
--- a/cli.js
+++ b/cli.js
@@ -521,7 +521,7 @@ process.on('uncaughtException', function(err) {
       break;
 
     case 'registry':
-      options = readOptions(args, ['yes']);
+      options = readOptions(args, ['yes', 'local']);
 
       if (options.yes)
         ui.useDefaults();
@@ -531,7 +531,7 @@ process.on('uncaughtException', function(err) {
       if (action === 'config') {
         if (!args[2])
           return ui.log('warn', 'You must provide an registry name to configure.');
-        return Promise.resolve(registry.configure(args[2]))
+        return Promise.resolve(registry.configure(args[2], options.local))
         .then(function() {
           ui.log('ok', 'Registry %' + args[2] + '% configured successfully.');
         }, function(err) {
@@ -543,7 +543,7 @@ process.on('uncaughtException', function(err) {
           return ui.log('warn', 'You must provide an registry name to create.');
         if (!args[3])
           return ui.log('warn', 'You must provide the registry module name to generate from.');
-        return Promise.resolve(registry.create(args[2], args[3]))
+        return Promise.resolve(registry.create(args[2], args[3], undefined, options.local))
         .then(function(created) {
           if (created)
             ui.log('ok', 'Enpoint %' + args[2] + '% created successfully.');
--- a/cli.js
+++ b/cli.js
@@ -437,9 +437,9 @@ process.on('uncaughtException', function(err) {
       if (options.globals)
         options.globalDeps = eval('(' + options.globals + ')');
 
-      if (options['global-defs']) 
+      if (options['global-defs'])
         options.globalDefs = eval('(' + options['global-defs'] + ')');
-      
+
       var bArgs = options.args.splice(1);
 
       if (bArgs.length === 0)
@@ -571,7 +571,7 @@ process.on('uncaughtException', function(err) {
 
     case 'c':
     case 'config':
-      options = readOptions(args, ['local'])
+      options = readOptions(args, ['local']);
       var property = options.args[1];
       var value = options.args.splice(2).join(' ');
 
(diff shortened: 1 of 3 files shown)


lib/global-config.js
```diff
@@ -115,25 +122,108 @@ if (exports.config.github) {
...
+    }
+    catch(e) {
+    if (e.code !== 'EEXIST')
+      throw 'Unable to create jspm system folder\n' + e.stack;
+    }
+  }
+
+  // If the user wants to save options to a local configuration file that is not
+  // created, we need to create it and extend it with the default config.
+  if ( saveLocal && !config.local.path ) {
```
guybedford: I'm just wondering about this logic, which will cause private data in the global config file to persist into the local project, potentially exposing passwords to version control by mistake.

guybedford: Sorry, I see this is just the default config being surfaced. I'm still not sure I understand why this is necessary though, if config is considered as the superposition of global and local configuration?

guybedford:
Thanks so much, will be great to get this in.

lib/global-config.js
hmps: Looking at it now, a month later, I share your confusion. Should work fine without it...

I'll make sure to test and confirm, and will update the PR if everything checks out!

hmps: Ok, found the reason why I added it... Without that if clause, if I try to save local config in a dir where no local config file exists, [L199](https://github.com/hmps/jspm-cli/blob/local-config/lib/global-config.js#L199) will throw because `config.local.path` is undefined.

The reason is that `config.local.path`, set on [L54](https://github.com/hmps/jspm-cli/blob/local-config/lib/global-config.js#L54) via `getLocalConfigPath()`, will be not be set if no local config file exists. So in the `if`, I `getDefaultLocalConfigPath()` to know where to save the generated config file.

That `config.local.path` is undefined is used to know which config to load ([L87](https://github.com/hmps/jspm-cli/blob/local-config/lib/global-config.js#L87)), that's why I don't set it to a default value in `getLocalConfigPath`.

Does that make sense to you @guybedford?

hmps:
@guybedford I'm the one who should be thankful! We use this every day at work and it is really working great for us!

lib/global-config.js
guybedford: Surely the default localConfigPath can be implied as the `path.resolve(config.pjsonPath, '.jspmrc')`?

guybedford:
Glad to hear that. I hope you don't mind, but because every line of code added is more future maintenance it will need to be very carefully reviewed to ensure code complexity is kept to a minimum.

hmps:
Don't mind at all, I'd be worried if you didn't scrutinize it :)

I guess what you're getting at is that `getDefaultLocalConfigPath` could be simplified? 

I'm not sure that I follow you on the `path.resolve(config.pjsonPath, '.jspmrc')` part though. If I run it in place of `getDefaulLocalConfigPath` I get [...]/jspm-test/package.json/.jspmrc.


On a side note, I did find that I used some pretty stupid variable names in registry.js. Will fix and push the changes.

hmps:
Thought of another thing...

If I create / config a registry and have a local config file, should the config be saved locally by default, or should it require the `--local` flag?

If locally by default, should there also be a `--global` flag?

I'm leaning towards the latter, it would make sense for local to be default if I have a default config available. But it should also possible to save globally from a locally configured project.

cli.js
```diff
@@ -543,7 +543,7 @@ process.on('uncaughtException', function(err) {
           return ui.log('warn', 'You must provide an registry name to create.');
         if (!args[3])
           return ui.log('warn', 'You must provide the registry module name to generate from.');
-        return Promise.resolve(registry.create(args[2], args[3]))
+        return Promise.resolve(registry.create(args[2], args[3], undefined, options.local))
```
guybedford: Let's alter the signature rather to `return Promise.resolve(registry.create(args[2], args[3], options.local))`

lib/registry.js
```diff
@@ -212,24 +223,33 @@ exports.configure = function(registry) {
...
   .then(function() {
-    globalConfig.save();
+    globalConfig.save(updatedConfig, setLocal);
   });
 };
 
 // jspm registry create mycompany jspm-github
 // creates a custom registry based on the given handler
-exports.create = function(name, handler, override) {
+exports.create = function(name, handler, override, setLocal) {
```
guybedford: `function(name, handler, local, override) {`

guybedford:
I'm thinking it might be nicer to use a getter / setter global API along the lines of:

```javascript
// automatically gives the config value from local or global as appropriate
globalConfig.get('some.property.value');
// local flag is optional
globalConfig.set('some.property.value', value, local);
```

We then ensure that all config operations from `registry.js` and `global-config.js` itself run through this getter / setter API, turning the internal `localConfig` and `globalConfig` objects into private data.

guybedford:
I think the default should always be to save to global config, unless `--local` is explicitly set. The reason for this is first and foremost backwards-compatibility as that is the existing behaviour. It is also effectively the same default npm uses.

brunopgalvao:
This would be great!

hmps:
The [changes in global-config.js](https://github.com/jspm/jspm-cli/commits/master/lib/global-config.js) made for a problematic merge. So instead of trying to figure out how to merge properly, in a codebase that I'm not 100% comfortable in, I'll close this PR and take another stab at it with the latest updates as base.

jabbera:
Did functionality to accomplish this ever get merged. I can't figure out how to build on our build server as we only allow outbound github downloads via our cache.

tjaartvdwalt:
Local config would be very handy for me in my CI environment.
Any chance of merging this code?

PR #1269, opened 2015-11-06
Honor existing line endings in config.js

stefanr:
If you are working with an OS that uses different line endings than your config.js, it's pretty annoying that these will be overwritten.

I think it would be nice to honor the existing line endings. This PR will also override the line endings in the package.json depending on the characters used in the config.js, but in my case this is also a nice side effect.

Probably the best approach would be to save the existing line endings somewhere in the corresponding configs (package, loader) and use them explicitly when writing the file.

### DIFF
--- a/lib/common.js
+++ b/lib/common.js
@@ -18,8 +18,7 @@ require('core-js/es6/string');
 var asp = require('rsvp').denodeify;
 var fs = require('graceful-fs');
 var path = require('path');
-var newLine = require('./config').newLine;
-var tab = require('./config').tab;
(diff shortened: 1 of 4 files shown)


stefanr:
Just recognized that this seems to be part of 0.17 ... so ignore this.

guybedford:
Thanks for checking!

guybedford:
Thanks for checking!

staxmanade:
Any thoughts on how can we leverage this? Not following how I can get my line endings to be `\n` instead of `\r\n`?

Issue #1274, opened 2015-11-09
Handle casing - ensure all package names are lowercased


guybedford:
By default we should just lowercase all package name inputs.

jackfranklin:
@guybedford would that be a case of lowercasing the inputs to https://github.com/jspm/jspm-cli/blob/master/lib/install.js#L168 ? I'm happy to work on the PR, but a pointer in the right direction would be appreciated to avoid wasted effort :)

guybedford:
@jackfranklin yes exactly, that should cover it all actually. Would be awesome to have a contribution here.

jackfranklin:
@guybedford perfect, PR incoming.

Am I right in saying that `name` is always a string, but `target` might be? 

EG, need to handle all these cases:

```
jspm.install('jquery', 'github:components/jquery@^2.0.0') => downcase both
jspm.install('jquery', {...}) => downcase just name
jspm.install('jquery', 'github:components/jquery@^2.0.0', {...}) => downcase first two
```

jackfranklin:
@guybedford this is now done as of #1275 being merged :tada:

guybedford:
I'm only just starting to think properly through the edge case upgrade path breaks on this now...

1. Registry entries at https://github.com/jspm/registry/tree/master/package-overrides/github would all need to be lower cased. Existing installs in jspm that are case sensitive would then break, so we probably want to duplicate the above entries into their lower case equivalents
2. If projects have previously installed a cased package, they would need to run `jspm install` to "downcase" their configuration file. Any System.import statements to the full name like `github:Project/name@x.y.z/...` would then need to be rewritten.

(1) is completely resolvable and just involves upgrading the registry.

(2) is a little bit more worrying though, and makes me think we may have to bump this change to jspm 0.17 as a breaking change unfortunately.

PR #1275, opened 2015-11-09
Downcase `name` and `target` inputs to jspm.install

jackfranklin:
To tackle #1274.

Not entirely sure how to write a test, so I didn't, but let me know @guybedford !

### DIFF
--- a/lib/config/package-name.js
+++ b/lib/config/package-name.js
@@ -26,6 +26,7 @@
   version: '2.0.3'
 */
 function Package(name, escape) {
+  name = name.toLowerCase();
   this.exactName = name;
 
   if (name.indexOf(':') !== -1)


guybedford:
Apologies, I forget that `target` is a PackageName class here (the comments are actually incorrect and should be removed, it is always an instance of this class).

We can probably just implement this as the first step in the PackageName class instantiation then - https://github.com/jspm/jspm-cli/blob/master/lib/config/package-name.js#L28.

jackfranklin:
Thanks @guybedford ! So we should still downcase `name`, but move the downcasing of `target` into `PackageName` ?

guybedford:
@jackfranklin I think we can skip downcasing `name` as it doesn't correspond to a file on the filesystem, which is where the casing conflicts occur.

jackfranklin:
@guybedford can you take a second look? Updated.

guybedford:
Looks great! :+1: to a +1 diff :)

PR #1291, opened 2015-11-15
Add support of non '.js' main

unional:
One custom step need is to add `packages: { "myModule": { defaultExtension: "{myExtension, e.g. ts}" } }`, which is ok to do before 0.17

### DIFF
--- a/lib/package.js
+++ b/lib/package.js
@@ -759,17 +759,28 @@ exports.createMain = function(pkg, pjson, downloadDir) {
       return true;
     }
 
-    if (main) {
-      if (main.startsWith('./'))
-        main = main.substr(2);
-      if (main.endsWith('.js'))
-        main = main.substr(0, main.length - 3);
+    if (!main)
+      return false;
+
+    if (main.startsWith('./')) {
+      main = main.substr(2);    
     }
+    
+    return new Promise(function(resolve) {
+      // Check to see if main points to an actual file or not
+      // i.e. does it include extension.
+      mainPath = path.resolve(downloadDir, main);
+      fs.exists(mainPath, resolve);
+    });
+  }).then(function(exists) {
+    if (exists)
+      return exists;
 
     main = main || 'index';
 
-    // try the package.json main
     return new Promise(function(resolve) {
+      // `main` does not have extension (or it does not point to an actual file).
+      // Try again with `.js` extension
       mainPath = path.resolve(downloadDir, main + '.js');
       fs.exists(mainPath, resolve);
     });


guybedford:
Thanks. Did you manage to take a look at the failing test?

unional:
Looking at it.

unional:
It actually fail on main. It's not related to this PR.  I checkout jspm/master and get same error.
Also validated the changed code is not being invoked in the failed test.

unional:
Really weird. Saw they pass on travis history, but I hard reset local to 0.16.14 (e5e9f27) and run test it still fail with same error.

guybedford:
@unional that would be due to caching. Run `jspm cc` and remove `testlibs/jspm_packages` before starting the tests.

unional:
Done!

lib/package.js
```diff
@@ -760,31 +761,50 @@ exports.createMain = function(pkg, pjson, downloadDir) {
...
+    
     main = lastNamePart;
 
-    if (main.endsWith('.js'))
-      main = main.substr(0, main.length - 3);
+    var match = /(\.[^.]+)$/.exec(main);
+    if (match) {
+      mainFileExtension = match[1];
+      main = main.slice(0, - mainFileExtension.length);
+    }
```
guybedford: I'm not sure I follow why this is necessary? Surely line 783 can be left as-is with the same effect?

lib/package.js
```diff
@@ -760,31 +761,50 @@ exports.createMain = function(pkg, pjson, downloadDir) {
...
+    });
+
+  }).then(function(exists) {
+    if (exists) {
+      var match = /(\.[^.]+)$/.exec(main);
+      if (match) {
+        mainFileExtension = match[1];
+        main = main.slice(0, - mainFileExtension.length);
+      }
+      return exists;
```
guybedford: Surely we only need to set `main` here and not `mainFileExtension` as well?

lib/package.js
```diff
@@ -760,31 +761,50 @@ exports.createMain = function(pkg, pjson, downloadDir) {
...
+      var match = /(\.[^.]+)$/.exec(main);
+      if (match) {
+        mainFileExtension = match[1];
+        main = main.slice(0, - mainFileExtension.length);
+      }
+      return exists;     
+    }
+
+    return new Promise(function(resolve) {      
+      mainPath = path.resolve(downloadDir, main + mainFileExtension);
```
guybedford: Again this seems redundant since the mainPath assignment in line 775 would be equivalent?

lib/package.js
unional: `mainFileExtension` is defaulted to `.js`. This part is new to check if the `pjson.main` is defined without extension. When the file does not exist, meaning that is the case here and we get the `mainFileExtension` out of it.

unional: I mean if the file does exist

lib/package.js
unional: This is checking if `pjson.main` does not point to a file directly, meaning the Default extension needs to apply.

lib/package.js
unional: There is an improvement to be made for the case that `main` was empty (thus setting it to `index`). The check wild be redundant there. 
Can improve it by defaulting to `index.js`, and skip the second check

lib/package.js
unional: This is checking for the `lastNamePart`.  Could it has non `.js` there? If it can't, then this is redundant.

lib/package.js
guybedford: I think the whole check can be simplified by just checking two things and hitting the existence path when the first matches - (1) if the main exists as written, (2) if the main exists with a default js extension added.

unional: It needs 3 checks:
1. `pjson.main` as is (may or may not have extension)
2. `pjson.main` + `.js` if (1) fails
3. `lastNamePart` from `pkg.name` if (1) and (2) fails.

The `lastNamePart` can be different than `main` in (1) and (2)., e.g.:
* `bower-traceur-runtime` vs `./traceur-runtime.js`
* `bower-traceur` vs './traceur.js`

unional: I can see that you want to reduce it to 2, i.e.
1. if `pjson.main` does not exists, try `pkg.name -> lastNamePart`. If not, try as is
2. If (1) fails and `pjson.main` exists, then try adding extension.
The flow is different, so I don't want to make too big a change (as I see you already refactored the whole thing in `0.17`).

Also, it will normally do no more than 2 check in current implementation anyway, because they will be skipped if the file is found.

If you want, I can do the refactoring, not much of a deal anyway. :)

unional:
Coverage test failed with:
Would you like to set up your GitHub credentials?:Yes
Enter your GitHub username:undefined

non-issue?

lib/package.js
```diff
@@ -760,24 +761,40 @@ exports.createMain = function(pkg, pjson, downloadDir) {
...
+        mainFileExtension = match[1];
+        main = main.slice(0, - mainFileExtension.length);
+      }
+      return exists;
+    }
 
-    // try the package.json main
     return new Promise(function(resolve) {
-      mainPath = path.resolve(downloadDir, main + '.js');
+      mainPath = path.resolve(downloadDir, main + mainFileExtension);
```
guybedford: This can just be `'.js'`.

lib/package.js
```diff
@@ -760,24 +761,40 @@ exports.createMain = function(pkg, pjson, downloadDir) {
...
-      if (main.startsWith('./'))
-        main = main.substr(2);
-      if (main.endsWith('.js'))
-        main = main.substr(0, main.length - 3);
+      if (main.startsWith('./')) {
+        main = main.substr(2);    
+      }
+    }
+    else {
+      main = 'index' + mainFileExtension;
```
guybedford: `'index.js'`

lib/package.js
```diff
@@ -745,6 +745,7 @@ exports.createMain = function(pkg, pjson, downloadDir) {
   var mainPath;
   var pluginMain;
 
+  var mainFileExtension = '.js';
```
guybedford: I don't think this variable is necessary?

lib/package.js
```diff
@@ -760,24 +761,40 @@ exports.createMain = function(pkg, pjson, downloadDir) {
...
+      mainPath = path.resolve(downloadDir, main);
+      fs.exists(mainPath, resolve);
+    });
+
+  }).then(function(exists) {
+    if (exists) {
+      var match = /(\.[^.]+)$/.exec(main);
+      if (match) {
+        mainFileExtension = match[1];
+        main = main.slice(0, - mainFileExtension.length);
```
guybedford: I'm not sure I understand this - why do you need to remove the extension at all?

lib/package.js
unional: Interesting. Let me look at it again.

lib/package.js
unional: That's what the original code does - removing `.js` from `main`. But as I read through the code, I agree with you. Keeping the extension will be a better solution, as the user can do:

````
packages: {
  "myModule": {
    defaultExtension: false
  }
}
````

which will be more aligned with 0.17, correct?

I'll make the change accordingly.

guybedford:
This looks good, thanks for getting it to this point!

Could you do a squash of these commits with a force push to clear up the history? Otherwise I can do it on merge no problem too.

unional:
Like this?

guybedford:
Looks like it caught some other stuff there, perhaps it needed a rebase as well?

unional:
Seems like the same error as the coverage failure last time.  Is there a way to retry the build without committing?

guybedford:
I'm not so worried about that, but see the extra file contents at https://github.com/jspm/jspm-cli/pull/1291/files.

unional:
Oh....how come...
Let me check.

unional:
I did revert+amend to fix it. Seems good now.  Is that the best way to do it? I still have lots of stuff to learn in using git.

And thanks for the careful code reviews. :)

guybedford:
:+1:

PR #1319, opened 2015-11-26
Avoid mutating process.argv

### LINKED ISSUE #1305: 'jspm run' - pass arguments to target script
I notice that command line arguments are not passed to the target script.  I would expect any arguments after the name of the target script to be treated as arguments to the script (as Node does).

The current behavior may be desirable, but I don't see an alternative mechanism to pass arguments to the target script.

jackfranklin:
This means that any script run with `jspm run` will get the regular
`process.argv`, and all args given to `jspm run` will make it through to
the target script. Closes #1305.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -205,7 +205,8 @@ process.on('uncaughtException', function(err) {
     return options;
   }
 
-  var args = process.argv.splice(2),
+  // [].concat() to avoid mutating the given process.argv
+  var args = process.argv.slice(2),
       options;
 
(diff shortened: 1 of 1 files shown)


guybedford:
Awesome, thanks so much Jack!

I think we can get this down by just replacing `splice` with `slice`?

jackfranklin:
@guybedford aha of course! Updated.

guybedford:
:+1:

PR #1320, opened 2015-11-26
Add `--production` flag to not install dev dependencies.

### LINKED ISSUE #1297: Support jspm install --production
Currently, running `jspm install` installs `dependencies` and `devDependencies`. This behavior was added in #757 to align it with `npm install`.

It would be awesome to be able to add the `--production` flag to exclude `devDependencies` from being installed.

jackfranklin:
Closes #1297.

@guybedford I'm not sure if it's as simple as this - but a quick test showed that this works fine for the base case. I feel like I might have missed something though.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -228,7 +228,7 @@ process.on('uncaughtException', function(err) {
     case 'i':
     case 'install':
       options = readOptions(args, ['force', 'link', 'yes', 'lock', 'latest',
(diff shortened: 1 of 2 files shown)


jackfranklin:
I also have no idea why this build randomly fails on Node 4. Think it's a weird error too :/

guybedford:
:+1:

PR #1325, opened 2015-11-27
Add #inspectVersions method to list installed packages

jackfranklin:
Returns a promise that resolves to an array of:

```
[
  { dep: 'npm:foo', version: '1.1.1', linked: false },
  ...
]
```

@guybedford how does this look? I still don't quite understand all the details of this so I might have missed something.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -776,7 +776,7 @@ function showInstallGraph(pkg) {
 exports.showInstallGraph = showInstallGraph;
 
 
-function showVersions(forks) {
+function getVersionsList() {
   installed = installed || config.loader;
 
   var versions = {};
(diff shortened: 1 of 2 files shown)


lib/install.js
```diff
@@ -805,7 +805,35 @@ function showVersions(forks) {
...
   });
 
-  versions = alphabetize(versions);
+  return {
+    versions: alphabetize(versions),
+    linkedVersions: linkedVersions
+  };
+}
+
+function inspectVersions(forks) {
```
jackfranklin: this `fork` argument actually isn't being used - not sure if I should be or not. A bit unclear on how this affects the output.

guybedford:
Thanks, glad to see the small diff principles in action!

We just need to consider the output data here a little bit carefully. Strictly, just taking the most recent version is incorrect because I could have installed a library to 1.0.1, while a dependency of mine uses version 3.5.0 of that same library. The version we're interested in is more likely to be the version I personally installed than the dependency.

The most general data structure to represent what I have installed is exactly the combination of map config and package map config - that is I have primary resolutions in the first map config, and secondary resolutions that trace out of the dependency maps from there.

Perhaps it would help to first understand your use case here a little better to see what this output is for? Then lets try to put something together that captures the tree metadata as generally as possible?

jackfranklin:
Sure :)

The main motivation is that I was working on a `jspm-outdated` script that would parse your package.json and list:

- the semver range for each jspm package
- the current installed version
- the latest version from github / npm

So that you could easily update and keep things up to that.

guybedford:
Perhaps then you could work with an output like:

```javascript
{
  "primary": {
    "bootstrap": {
      "target": "github:twbs/bootstrap@^3.0.0",
      "resolution": "github:twbs/bootstrap@3.0.1",
      "linked": false
    }
  },
  "secondary": {
    "github:twbs/bootstrap@3.0.1": {
      "jquery": {
        "target": "github:components/jquery@^2.0.1",
        "resolution": "github:components/jquery@2.3.4",
        "linked": false
      }
    }
  }
}
```

Where users have added custom maps, the resolution might be a local map like `./local/path.js` in those cases.

Let me know your thoughts on the above?

jackfranklin:
That looks pretty good :+1:

guybedford:
That this would likely no longer fall under the `jspm inspect` function though. It may just be an API function for analysis, perhaps along the lines of `getInstallTree` or something similar.

jackfranklin:
That's fine - this only started out being under `inspect` because I was stealing the code ! I'm happy for it to go wherever you think makes sense :)

PR #1332, opened 2015-11-30
Git book for JSPM

OrKoN:
The book looks like this: http://blog.alex-rudenko.com/jspm-cli/index.html
What do you think?  Would be great to put it on jspm's website and have everything published as a book (maybe, together with systemjs, builder, jspm). Or did you have other plans for documentation in mind?


guybedford:
Nice! I've created https://github.com/jspm/jspm.github.io so it should be possible to push this there.

PR #1333, opened 2015-11-30
A page about CLI params

OrKoN:
Listed params that I am asked quite often

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -1,0 +1,37 @@
+## Getting Help
+
+```sh
+jspm --help
+```
+
+## Bundles
+
+```sh
+jspm bundle moduleA + module/b [outfile] [flags]
+```
+
+| Flag                        | Description  | JS API Equivalent|
+| ----------------------------|---------------| -------------------| 
+| `--minify`                    | minify the bundle using uglify-js | `minify` |
+| `--no-mangle`                 | applies if --minify is true, mangle the source code or node | `mangle`
+| `--inject`                    | whether to inject the bundle into config.js | `inject`
+| `--skip-source-maps`          | skips source maps generation | `sourceMaps`
+| `--source-map-contents`       | adds `sourcesContent` into the source map | `sourceMapContents`
+
+## SFX Bundles
+
+```sh
+jspm bundle-sfx moduleA + module/b [outfile] [flags]
+```
+
+| Flag                        | Description  | JS API Equivalent|
+| ----------------------------|---------------| -------------------| 
+| `--minify`                    | minify the bundle using uglify-js | `minify` |
+| `--format`                    | sfx module format <amd|cjs|global> | `format` 
+
+## Global Flags
+
+| Flag                        | Description  |
+| ----------------------------|---------------|
+| --yes &#124; -y                 | Skip prompts / use default inputs |
+| --log <ok&#124;warn&#124;err>       |Set log level|


guybedford:
Thanks this and https://github.com/jspm/jspm-cli/pull/1334 are a great start - I will leave this up until I can comprehensively go through this.

The gitbook is now on the live site though.

OrKoN:
Thanks! Actually, after I started writing this page I understood that it is pretty useless. I have another idea: another PR that uses commander.js to manage CLI params and extend it with the generation of Markdown. What do you think? I am only wondering if you considered using commander.js and what stopped your from doing so?

guybedford:
The reason commander is not used is because we needed the ability for flags to be greedy. For example when doing override flags, we need `--override "{...}"` to be able to contain arbitrary syntax. Plus command line arg parsing code really isn't much work at all, so there's no reason not to use a custom solution for more control.

In terms of the docs, this is where I get opinionated. I'm somewhat against automated documentation tools or heavy documentation structure, as I think documentation should be written for humans, and such be completely customized. Each command should be documented in the way that makes the most sense for it, without having to rely on preset patterns of command / flag documentation slots that may or may not make sense for the given command.

OrKoN:
> In terms of the docs, this is where I get opinionated. I'm somewhat against automated documentation tools or heavy documentation structure, as I think documentation should be written for humans, and such be completely customized. Each command should be documented in the way that makes the most sense for it, without having to rely on preset patterns of command / flag documentation slots that may or may not make sense for the given command.

I agree with this but there is one problem: the documentation should be up-to-date and w/o automation it's hard to keep up. I have noticed that many people have troubles finding the right CLI options or explanation of how it works. In my opinion, it would be great if human-targeted documentation was embedded in the source code and used as a single source to re-construct the usage details and online documentation. 

For example, one of the common things that people miss is minify/mangle flags, --skip-source-maps and bundle arithmetic. And it's not possible to find `--skip-source-maps` in the docs. So I feel like a CLI/API reference is definitely needed. And from my experience, many people don't want to read all docs available in the repo or guides: normally, they know what they are looking for based on their experience or common sense, for example `how to I disable source maps` but have troubles finding the exact syntax.

I think commander supports `--override`-like patterns and it also supports variadic arguments.  Let me explore this path a bit and if it can work (with or without commander) I will let you know.

guybedford:
I completely agree that the flag documentation is lacking, and the `jspm` help output is not enough to cover everything. This will be a high priority once jspm 0.17 alpha is out.

Command-level help via eg `jspm install --help` or similar would be great, and I don't think it is a technical problem in terms of how to structure this, but rather one of documentation content to fill these spaces.

OrKoN:
As an experiment I have tried to implement a small library that helps with documenting CLI params and supports rendering to the console and Markdown. It will be able to assist with command-level help as well. And it does not try to automate parsing of arguments at all (compared to commander).

I have got some screenshots:

CLI:
selection_196

Markdown:
selection_197

I am still working on a better way to model properties and on a better output (especially for MD, it's not so nice at the moment). Let me know what your opinion about this is.

guybedford:
@OrKoN this looks really interesting. I like the idea of having a single markdown doc as the primary source here.

In terms of how the CLI system parses that markdown, perhaps we just need a generic way of converting markdown to terminal output and apart from parsing out the headings based on command names, leave it at that?

Perhaps using something like https://www.npmjs.com/package/marked-terminal would help here if you aren't already?

OrKoN:
Sorry, it seems I will not be able to finish this. Closing this for now..

guybedford:
Sure I understand, thanks for the update.

PR #1336, opened 2015-12-01
Improved documentation for GitHub Authentication Environment Variable

joeldenning:


guybedford:
Thanks, this may be something more suitable for other resources than the README though as it is unix specific and not generally encouraged for most users to generate this themselves.

joeldenning:
Fair enough, those commands are definitely unix specific.

Although we found it useful to generate the authToken on our own since we use a different auth for our CI than we do for our local development environments. Is there a better way to get JSPM to prompt you for username and password (in order to generate JSPM_GITHUB_AUTH_TOKEN) without having to go through the process of getting rate limited by Github? That was the tedious part that we were wanting to skip by generating the base 64 token on our own. If there is a way to do that, is that documented?

guybedford:
One way to handle it might be to allow command line arguments to be provided along with `jspm registry config github` and then have this under a `jspm registry config github authtoken` option or something similar.

I think this problem will become less important in jspm 0.17 though as it will no longer depend on github credentials, but it will continue to apply for any custom registry auth though.

aluanhaddad:
:+1: for allowing command line arguments to be provided along with ```jspm registry config github```

Issue #1198, opened 2015-10-12
Intermittent issues when bundling: unable to load registry


OliverJAsh:
Very occasionally when we bundle, we get the following error:

```
[16:59:39][Step 2/4] /home/ubuntu/buildAgent/work/4541e69dce9eb322/node_modules/jspm/lib/registry.js:186
[16:59:39][Step 2/4]     throw 'Unable to load registry %' + registry + '%';
[16:59:39][Step 2/4]                                                  ^
[16:59:39][Step 2/4] Unable to load registry %bower%
```

`jspm bundle` seems to mutate the jspm config file to invalid JSON (note the missing comma):

```
{
  "defaultTranspiler": "babel",
  "defaultRegistry": "jspm",
  "strictSSL": true,
  "registries": {
    "github": {
      "handler": "jspm-github",
      "remote": "https://github.jspm.io"
    },
    "npm": {
      "handler": "jspm-npm",
      "remote": "https://npm.jspm.io"
    },
    "jspm": {
      "handler": "jspm-registry",
      "remote": "https://registry.jspm.io"
    }
  }
}"bower": {
      "handler": "jspm-bower-endpoint"
    }
  }
}
```

Reconfiguring the registry corrects the config file.

Any pointers to where jspm bundle is mutating the config? I will take a look :smile:

guybedford:
This usually happens due to multiple instances of jspm running at the same time together. I'm actually looking at better configuration locking for jspm 0.17. If you want to look into a patch that is welcome too.

guybedford:
(I believe it's an actual file system lock that is needed)

OliverJAsh:
@guybedford Can you point me to where the write is happening? I might have a look into this.

guybedford:
@OliverJAsh sure, that would be great! The global config save function is at https://github.com/jspm/jspm-cli/blob/master/lib/global-config.js#L76. This is currently sync but usage could be upgraded to be async. It gets called within that file itself and then at https://github.com/jspm/jspm-cli/blob/master/lib/registry.js#L218.

The global config load does need to be sync though as that enables sync usage of builder class instantation for example.

guybedford:
This should be fixed with the config lock in https://github.com/jspm/jspm-cli/commit/fa03ff55a5793ec9e760bbeab686e5a9c24d3ae5.

guybedford:
(pending release)

Issue #1076, opened 2015-08-30
Unable to read global configuration file (invalid JSON in .jspm/config)


Mosho1:
This is my global `.jspm/config`:

    {
      "defaultTranspiler": "babel",
      "defaultRegistry": "jspm",
      "strictSSL": true,
      "registries": {
        "github": {
          "handler": "jspm-github",
          "remote": "https://github.jspm.io"
        },
        "npm": {
          "handler": "jspm-npm",
          "remote": "https://npm.jspm.io"
        },
        "jspm": {
          "handler": "jspm-registry",
          "remote": "https://registry.jspm.io"
        }
      }
    }     "remote": "https://npm.jspm.io"
        },
        "jspm": {
          "timeouts": {
            "lookup": 60,
            "build": 120
          },
          "handler": "jspm-registry",
          "remote": "https://registry.jspm.io"
        }
      }
    }

Which is invalid. Not sure what I did today that tried to update this. Using 0.16.0

guybedford:
This can happen if you ran two instances of jspm simultaneously, we're not currently set up for write locking the config file.

PR #1347, opened 2015-12-09
Add file locking for global config

apiv:
This pull request should solve issues file locking issues raised in #1198 and #1076.

Basically, I rearranged the global-config.js file to do this, in order:

   1) lock ~/.jspm/config
   2) read ~/.jspm/config
   3) apply the defaults/migrations with `dprepend`
   4) save the config _only_ if it is different from the file on disk.
   5) unlock the file

This allows multiple jspm processes to start/run concurrently, without accidentally overwriting the global config when another process is writing to it.

### DIFF
--- a/lib/global-config.js
+++ b/lib/global-config.js
@@ -18,7 +18,10 @@ var ui = require('./ui');
 var fs = require('graceful-fs');
 var path = require('path');
 var HOME = require('./config').HOME;
+var lockFile = require('proper-lockfile');
 var dprepend = require('./common').dprepend;
+var readJSONSync = require('./common').readJSONSync;
+var stringify = require('./common').stringify;
 
 // global config - automatically created and loaded on startup
 exports.registries = [];
@@ -43,38 +46,11 @@ if (process.env.USERPROFILE && HOME !== process.env.USERPROFILE && !fs.existsSyn
   }
 }
 
-function save() {
-  try {
-    fs.mkdirSync(HOME + path.sep + '.jspm');
-  }
-  catch(e) {
-    if (e.code !== 'EEXIST')
-      throw 'Unable to create jspm system folder\n' + e.stack;
-  }
-  try {
-    fs.writeFileSync(globalConfigFile, JSON.stringify(exports.config, null, 2));
-  }
-  catch(e) {
-    throw 'Unable to write global configuration file\n' + e.stack;
-  }
-}
-if (fs.existsSync(globalConfigFile)) {
-  try {
-    exports.config = JSON.parse(fs.readFileSync(globalConfigFile).toString() || '{}');
-  }
-  catch(e) {
-    exports.config = {};
-    ui.log('err', 'Unable to read global configuration file.\nEnsure you have read and write access to %' + globalConfigFile + '%.');
-    process.exit(1);
-  }
-}
-else {
-  exports.config = {};
-  if (HOME)
-    save();
-}
-exports.save = save;
+// Begin file lock
+lock();
 
+// after loading, we'll dprepend all defaults to `exports.config` in this `try` block
+exports.config = readJSONSync(globalConfigFile);
 
 // NB can deprecate with jspm 0.14
 if (!exports.config.defaultRegistry && exports.config.registry)
@@ -86,9 +62,14 @@ if (!exports.config.registries && exports.config.endpoints)
 // delete exports.config.registry;
 // delete exports.config.endpoints;
 
-/*
- * Populate default registry configuration
- */
+// config upgrade paths
+// NB can deprecate with jspm < 10
+if (exports.config.github) {
+  dprepend(exports.config.registries.github, exports.config.github);
+  delete exports.config.github;
+}
+
+// populate default registry configuration
 dprepend(exports.config, {
   defaultTranspiler: 'babel',
   defaultRegistry: 'jspm',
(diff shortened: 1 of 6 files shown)


guybedford:
Thanks very much for this.

I'm just trying to figure out how this deals with the cases here. Could you perhaps help my understanding by answering some of the following questions?

* What happens if we have two processes, one running for 5 minutes, and another that runs for 2 minutes. The 2 minute process starts, reads the config. Then a minute later the 5 minute process starts and does the same. After a minute, the 2 minute process saves. Then after another 4 minutes the 5 minute process saves. Am I right in thinking the 5 minute process config would still be the one to persist, and this has nothing to do with the level the lock file acts at?
* If two processes run at the same time, and attempt to save to the config file at the same time, what happens to the second process when it hits the lock? Does this throw an error, or stall until the other process has finished, and then overwrite the file with the new data?

Do you think we need to include https://github.com/IndigoUnited/node-proper-lockfile#graceful-exit here for exits? Or does the latest Node handle these cases now?

apiv:
No problem! Looking forward to getting this in, as it would speed up our build significantly. Ran in series, our bundles will build in about 5 minutes. Ran in parallel, our entire build finishes in just 34 seconds. Being able to do this safely would be awesome (until now we had just kept our fingers crossed and kept backups of our config files in case they are written).

So, based on your questions, here are a few clarifications that I think will help explain the process:
- The lock does not exist for the entirety of the process. The file will only be locked on initial script run, as it reads and saves the config file. After it calls save, the file is unlocked, and another process can access it.
- If a lock cannot be acquired initially, it will attempt to retry up to 10 times at random intervals.

Before this refactor, the reads/writes on ~/.jspm/config for multiple processes would look like:
```
| PROC #1: read, write
| PROC #2: read, write
| PROC #3: read, write
| PROC #4: read, write
▼ PROC #5: read, write
```

This is inefficient, since all of these processes will be writing the same data. And of course, we have the issue where if one process attempts to read the file as another process is writing, it will fail (causing it to try to write the file, perpetuating the issue), which looks something like this:

```
| PROC #1: read, write
| PROC #2: cannot read while proc 1 is writing, write
| PROC #3: cannot read while proc 2 is writing, write
| PROC #4: cannot read while proc 3 is writing, cannot write while proc 3 is writing
▼ PROC #5: cannot read while proc 3 is writing, cannot write while proc 3 is writing
```

When _n_ identical concurrent processes are running, we only want to read once, write once, and then read from then on. Like so:

```
| PROC #1: lock, read, write, unlock   PROC #2: attempt to lock, fail, retry in 232ms   PROC #3: attempt to lock, fail, retry in 287ms
| PROC #2: lock, read, unlock          PROC #5: attempt to lock, fail, retry in 202ms
| PROC #3: lock, read, unlock          PROC #5: attempt to lock, fail, retry in 402ms
| PROC #4: lock, read, unlock
▼ PROC #5: lock, read, unlock
```

I tried to pepper in some failure to lock / retry scenarios in there, to explain how the process works.

So to answer your questions:

1) I'll label your 2-minute process **`PROC #1`**, and your 5-minute process **`PROC #2`**.
   ```
   T+0:00 PROC #1: start
   T+0:00 PROC #1: lock
   T+0:01 PROC #1: read     # the file is read, and the defaults are applied
   T+0:04 PROC #1: write    # with the defaults applied, the version in mem is different from disk, so it is saved
   T+0:05 PROC #1: unlock
   T+1:00 PROC #2: start
   T+1:00 PROC #2: lock
   T+1:01 PROC #2: read     # the file is read (this is the version that PROC #1 saved)
   T+1:02 PROC #2: unlock   # with the defaults applied, the version in mem is identical to disk, so it is not saved
   T+2:00 PROC #1: stop
   T+5:00 PROC #2: stop
   ```

2) So if another process cannot acquire a lock right away, it will retry again at a random interval, up to 10 times. A process failing to lock on this file in 10 tries is pretty statistically impossible, since the only real conflict occurs when the first `write` is happening. But, we do have a handler for that scenario. If it cannot acquire a lock after 10 tries, it will throw an error `Unable to obtain a lock`, and will not do anything further.

3) I'm doing some research now on SIGINT and SIGTERM and graceful exiting in node. What is the lowest version of node that JSPM tries to support? (A few of our projects here still run on 0.12.7).

apiv:
So after some research, it looks like we'll want to include these event handlers for windows environments. [The node docs state that only non-windows hosts have default handlers for these](https://nodejs.org/api/process.html#process_signal_events). In that case, we definitely want the lockfile to be removed.

I've pushed up a new, rebased commit with this change.

lib/global-config.js
```diff
@@ -43,80 +46,89 @@ if (process.env.USERPROFILE && HOME !== process.env.USERPROFILE && !fs.existsSyn
...
+      }
+    }
+  });
+
+  if (HOME)
+    save();
+}
+catch (e) {
+  if (e.code === 'ELOCKED')
+    throw 'Unable to lock global config file, not overwriting';
```
guybedford: I think we can remove this try-catch wrapper if we move this error check into the `lock` function itself?

lib/global-config.js
```diff
@@ -43,80 +46,89 @@ if (process.env.USERPROFILE && HOME !== process.env.USERPROFILE && !fs.existsSyn
...
+}
+finally {
+  unlock();
```
guybedford: This would need to move into the `save` function?

guybedford: Actually shouldn't this `unlock` come before running the `if (HOME) save()` step above?

guybedford:
Thanks for the explanations. So we still have the potential of config conflicts if I run say `jspm config x y` while a long-running process is going, but we avoid configuration corruption which is the primary feature here.

This is great to me, and I'm happy to merge. Just let me know your thoughts on the code feedback above.

lib/global-config.js
apiv: I can definitely move the try...catch with error handling into the lock function.

lib/global-config.js
apiv: Well the whole idea is to keep a single lock throughout the entire process. The lock is only released on three explicit conditions:

1) The process is killed, the lock is released on the exit event
2) An error is thrown, the finally block unlocks the file
3) The procedure completes successfully

So I wouldn't want to `unlock` the file here. I'll explain a bit more in my last comment.

guybedford: But then `save` should not be creating a lock?

apiv: Actually, once I move the try..catch to the `lock` function, it would make sense to move this into the `save` function. So I'll do that.

apiv:
Changes have been pushed.

guybedford:
This is great, thanks! I'm aiming to get a release out very soon.

ArmorDarks:
Strangely enough, I'm still getting a similar error when running two JSPM instances:

```
err  Unable to write global configuration file
```

It doesn't happen locally, on my Window machine, but does _sometimes_ on CircleCI Linux machine.

What is more strange, files aren't changing, so following description in this PR, JSPM shouldn't even try to write them, but error states that JSPM unable to write config file.

guybedford:
@ArmorDarks that is unfortunate to hear. If you are able to track this down further a PR would be very welcome.

PR #1355, opened 2015-12-11
Allow boolean and integer types with jspm config CLI

apiv:
One more small code change for your review. Right now, `jspm config strictSSL false` will set the value of the `strictSSL` key in `~/.jspm/config` to string `"false"`, instead of boolean false. Same for integers (eg registries.github.maxRepoSize).

This code change will read "true" and "false" from the command line and convert them to booleans, and convert strings of numbers to integers.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -205,6 +205,17 @@ process.on('uncaughtException', function(err) {
     return options;
   }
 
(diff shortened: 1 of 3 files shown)


guybedford:
Excellent, thank you!

PR #1403, opened 2016-01-07
Simple override example

peteruithoven:
The only example of the cli override syntax I know if is found in a sub doc page of the jspm/registry: 
https://github.com/jspm/registry/wiki/Configuring-Packages-for-jspm#testing-configuration


guybedford:
Sure, thanks for the suggestion.

PR #1449, opened 2016-01-20
When linking use `*` as pkg name format by default

gcnm-pgeil:
See https://github.com/jspm/jspm-cli/issues/1450


guybedford:
:+1:

PR #1463, opened 2016-01-24
Find linked package root to read package.json, fixes #1462

### LINKED ISSUE #1462: 0.17 linking and package.json derivation
I got the current branch from b49f8ec196decd3b92311c75bc589e33a8875d64 and while linking in several packages I noticed there is a bug with the way the meta json information for linked packages is computed.

All of my packages make use of the `directories.dist` property which took a while for me to realize now needed to be at the top level of package.json rather than inside `jspm`. It appears to me that `package.json` is [read from the link install target](https://github.com/jspm/jspm-cli/blob/b49f8ec196decd3b92311c75bc589e33a8875d64/lib/package.js#L523) instead of the link source. Thus the configuration is an empty slate for computing main.

cehoffman:
There might be a preference for a better error message, but that error condition should be nearly non existant. Only case is if a user deletes the linked package.json after sometime and then relinks the package.

### DIFF
--- a/lib/package.js
+++ b/lib/package.js
@@ -520,7 +520,12 @@ exports.download = function(pkg, options, installDeps, installPeerDeps) {
       if (options.linked && !options.unlink) {
         ui.log('info', 'Processing linked package `' + pkg.exactName + '`');
 
-        return Promise.resolve(readJSON(path.resolve(downloadDir, 'package.json')))
+        return asp(fs.realpath)(downloadDir)
+        .then(function(dir) {
+          return findPackageRoot(pkg, dir);
+        }).then(function(root) {
+          return readJSON(path.resolve(root, 'package.json'));
+        })
         .then(function(packageConfig) {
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks! This looks great, the only thing that worries me is if there either happened to be a `package.json` file inbetween the linked package base folder and the dist folder, or if there wasn't any package.json at all and we end up backtracking below the linked package.

I think we should try and come up with an approach that can somehow guard against those cases but nothing springs to mind. Will put some thought to it as well.

guybedford:
Actually I think this is pretty straightforward. We can guard against these cases when running the link creation initially:

If the link task sees there is a dist folder, it then runs this algorithm to ensure that the package.json picked up is the base-level package.json. If not it then throws the entire linking process complaining that there is package.json in-between, and the link should either be run from that folder or that package.json should be removed.

guybedford:
:+1: I see you've actually covered the root case perfectly!

PR #1606, opened 2016-03-11
0.17: Change "jspm.js" to "jspm.config.js" in upgrade notes

bendavis78:
Got a little confused when upgrading to 0.17, as there was no file named `jspm.js`. I'm assuming this should read `jspm.config.js`.


guybedford:
Thanks!

PR #1617, opened 2016-03-17
Fix skip encode names

AdamBurgess:
Alternatively this and skip-rollup could be moved to not even need an if statement (similar to `no-mangle`), which looks a little cleaner


guybedford:
Ha, thank you for spotting that!

PR #1618, opened 2016-03-17
Fix build errors not exiting with 1

AdamBurgess:
I noticed my CI builds were completing even though the build had an error, this should fix it.
The error thrown here is caught up the pipeline in cli.js where it correctly `process.exit(1)`'s.


guybedford:
:+1:

PR #1632, opened 2016-03-20
Init selection option

MSakamaki:
Small but I tried to add functionality to the branch 0.17.

jspm is a feature that it is possible to switch the parameter options is specified in the key of up and down at the time of the init.

I will let me send a pull request because I like this feature in the init process of karma.

### DIFF
--- a/lib/ui.js
+++ b/lib/ui.js
@@ -44,6 +44,41 @@ MutableStdout.prototype._write = function(chunk, encoding, callback) {
   callback();
 };
 
+
+var CtrlOption = function(options, message, def){
+  this.options = options || {};
+  this.questionMessage = message || '';
+  this.def = def || '';
+};
+
+CtrlOption.prototype = {
+  cursol: 0,
+  moveUp: function(){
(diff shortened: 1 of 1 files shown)


guybedford:
Nice, what was the reason for closing this?

MSakamaki:
I'm sorry embarrassing, 
but I have put a pull request without looking at the [contribute guidelines](https://github.com/jspm/jspm-cli/wiki/Contributors%27-Guide).

https://waffle.io/jspm/jspm-cli

make a issues to waffle, I was thinking throw a request with the correct procedure.

Sorry for the confusion

guybedford:
Not to worry, we don't actually use waffle anymore that much, the important thing though is to discuss the implementation details a little. Since you've started already though, happy to run the feedback on the PR especially since its such a small well-defined piece of code.

MSakamaki:
thank you for answer.
right away, let re-open again :)

guybedford:
This works really great! Merged manually for the next release.

PR #1654, opened 2016-03-29
Watch the current working directory instead of the lowest directory

oligot:
I just tested the new watch option in a project where the entrypoint
is in a subdirectory (app/app.js) and files changed were not detected
when running the command
> jspm bundle app/app.js dev-bundle.js --watch

By using the cwd, this now works as expected

### DIFF
--- a/lib/bundle.js
+++ b/lib/bundle.js
@@ -116,7 +116,7 @@ exports.depCache = function(expression) {
   var systemBuilder = new Builder();
 
   expression = expression || config.loader.main;
-  
+
   ui.log('info', 'Injecting the traced dependency tree for `' + expression + '`...');
 
   return systemBuilder.trace(expression, { browser: true })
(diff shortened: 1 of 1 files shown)


guybedford:
The `lowestDirectory` option is a little more correct here as jspm can run in a subpath. Can you perhaps confirm what the value of `lowestDirectory` was in your case and why it wasn't working correctly?

guybedford:
And thanks for testing this out!

oligot:
The value of `lowestDirectory` is _app_ in my case.
It's not working as it's trying to watch files (for ex: _app/app.js_) inside the app directory (instead of the cwd), thus trying to watch the file _app/app/app.js_ which doesn't exist.

Maybe we should remove the path `lowestDirectory` from the files before passing them as a glob option to sane ?

guybedford:
Thanks for clarifying. I've added a fix in https://github.com/jspm/jspm-cli/commit/726a39c008be1297d66d3fe2b322c6357b96484c.

oligot:
It now works as expected, thanks !

PR #1656, opened 2016-03-29
watch: reconstruct the full file path when calling the watch callback

oligot:
Otherwise, the file is not invalidated (as it points to a file that
doesn't exist), so the rebuild doesn't do anyting and the build file
is not up to date


guybedford:
Ah of course! Thank you :+1:

PR #1658, opened 2016-03-29
Use Facebook watchman for better compatibility. Especially BSD's

mgamsjager:
Nodejs fs watcher (which is the default) does not work on FreeBSD Unix. Running sane in watchman mode does result in triggering rebuild.

Other options would be Chokidar which also works reliable on FreeBSD.

### DIFF
--- a/lib/bundle.js
+++ b/lib/bundle.js
@@ -271,7 +271,7 @@ function buildWatch(files, fn) {
     return file.substr(lowestDir.length + 1);
   });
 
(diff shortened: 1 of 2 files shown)


guybedford:
Thanks for letting me know. I've implemented a fallback-based approach in https://github.com/jspm/jspm-cli/commit/eb72e2859a04c0c9361ab3ca21fa6508d9fe315e.

PR #1694, opened 2016-04-07
watch mode: support symlinks (fix #1687)

### LINKED ISSUE #1687: jspm bundle -wid with linked packages?
Is there a way to use 

> jspm bundle src/app.js -wid

and have it work with linked packages as well? 
It picks up everything in /src but not changes to the linked packages. I'm not using hot-reloading, just jspm link.

If it doesn't work right now, are there plans to support this, because that would be super cool.

oligot:
sane doesn't support symlinks[1].
Let's switch back to chokidar which supports symlinks.

Only tested on Linux: should probably test on Windows and OS X
to see if it works as expected.

[1] https://github.com/amasad/sane/issues/42


guybedford:
Thanks, I really like the approach in sane though a lot, and much more than chokidar. I was thinking of adding manual support for package symlinks by reading the package directory listings and statting them at build time.

guybedford:
https://github.com/jspm/jspm-cli/blob/master/lib/install.js#L931 already exists to read all the package folder paths from jspm_packages.

oligot:
Ok.
I'll let you change the code then.

guybedford:
I'm wondering if the best way to support symlinks might not be to just make a PR directly to sane itself here?

guybedford:
//cc @ebpa for the last comment.

guybedford:
On second thoughts that is a much more complex problem. Restricting symlinks to package folders in jspm_packages gives us a reasonable simplification of the problem to make it doable.

guybedford:
Closing for lack of activity. Feel free to reopen.

PR #1716, opened 2016-04-13
If error on init or inspect, exit with error code 1

AdamBurgess:
similar to my previous change, log the error but also quit with code 1.


guybedford:
👍

Issue #1718, opened 2016-04-13
0.17 -- rollup messes up the watcher when doing `jspm build -w`


joeldenning:
The impact is that only some files are actually watched for changes. The reason is that [`output.modules`](https://github.com/jspm/jspm-cli/blob/0.17/lib/bundle.js#L266) from systemjs builder doesn't include the modules that were inlined by rollup.

guybedford:
Thanks, but this should be fixed in SystemJS builder itself in https://github.com/systemjs/builder/blob/master/lib/builder.js#L649 by returning the full tree modules instead of the compiled modules. I can try to get to this when I can if it's not clear there.

guybedford:
I've added https://github.com/systemjs/builder/commit/dd3dcdf35fca8752245591ac6ab3e1743aeb7711 which should fix it here. If you're able to test that as well to verify that would be a great help.

joeldenning:
Verifying it now locally.

joeldenning:
Just verified that the fix works. The saw the watcher fail to pick up a change in an inlined module, then I plopped https://github.com/systemjs/builder/blob/dd3dcdf35fca8752245591ac6ab3e1743aeb7711/lib/builder.js into my node_modules and then the watcher did work correctly. Thanks for the quick turnaround on this.

guybedford:
Released in 0.17.0-beta.14.

PR #1719, opened 2016-04-13
Fixing #1718 - watcher now watches files inlined by rollup.

joeldenning:
See #1718 for the explanation -- basically `output.modules` doesn't include the full list of files to watch because of rollup inlining.


guybedford:
Let's continue to track implementation status in https://github.com/jspm/jspm-cli/issues/1718.

PR #1734, opened 2016-04-19
Close watcher immediatly

samouss:
When the file changes quickly, the callback function is executed more than once if the build time is too long. 

So the watcher is added multiple times, because he's closed after the build.

It just need to close it immediately for avoid multiple call.

### DIFF
--- a/lib/bundle.js
+++ b/lib/bundle.js
@@ -314,10 +314,8 @@ function buildWatch(files, fn) {
   watcher.on('add', function(filepath) {
     var file = path.join(lowestDir, filepath);
     ui.log('ok', 'File `' + file + '` created, rebuilding...');
-    fn(file, true)
-    .then(function() {
-      watcher.close();
-    });
+    fn(file, true);
+    watcher.close();
   });
   watcher.on('delete', function(filepath) {
     var file = path.join(lowestDir, filepath);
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks! This is great, but we shouldn't miss change events during the gap, rather we should queue up some state that notes a file was changed during the rebuild and we need to do another rebuild.

samouss:
Thanks for your answser, I'll check for another solution with cache or something like that.

guybedford:
It should be possible with just a boolean variable that tracks whether a "rebuild" is necessary.

samouss:
I changed a little bit the implementation, now the watcher is added only once on the first build.

A file change always trigger a rebuild, and we keep the original watcher for all next build. 

This avoid to add multiple watchers when the build is a bit long.

guybedford:
Thanks, it makes sense to try and minimise rewatching, but at the same time, we recrawl the full tree to detect any newly added files (which won't get watched if we just retain the existing watcher).

Ideally the watcher could be shared when the list of files hasn't changed, and then only recreated when it does, but you don't have to worry about this complexity in the implementation if it is too much for this PR.

joeldenning:
I was just about to make an issue for this but am even happier to see this pull request that fixes it. Thanks!

guybedford:
I went ahead and integrated this in https://github.com/jspm/jspm-cli/commit/504c27b4122d1d39a5f2c85995fbeead3b6476a5. Thanks for pushing this forward.

samouss:
Thanks !

PR #1737, opened 2016-04-20
File watching: Use sort to select the lowest directory for `buildWatch`

nickL:
### File watching: Use sort to select the lowest directory for `buildWatch`

Really cool to see the new file watching in `0.17` @guybedford , excellent job on this lib.

Small fix: I noticed that the `buildWatch` fn sometimes selects the lowest directory incorrectly.  It now loops through the entire module tree and as a result some `jspm_packages` can resolve as "lower" than the main application code.  Sorting the files by their length first  then selecting the shortest directory seems to fix it. 

LMK what you think. Happy to help debug/test further.

### DIFF
--- a/lib/bundle.js
+++ b/lib/bundle.js
@@ -285,11 +285,10 @@ function buildWatch(files, fn) {
   var sane = require('sane');
 
   // get the lowest directory from the files listing
-  var lowestDir = path.dirname(files[0]);
-  files.forEach(function(file) {
-    if (path.dirname(file).length < lowestDir.length)
-      lowestDir = path.dirname(file);
-  });
+  files.sort(function(f1,f2){
+    return path.dirname(f1).length -  path.dirname(f2).length
+  })
+  var lowestDir = path.dirname(files[0])
 
   var relFiles = files.map(function(file) {
     if (lowestDir == '.')
--- a/lib/bundle.js
+++ b/lib/bundle.js
@@ -285,10 +285,11 @@ function buildWatch(files, fn) {
   var sane = require('sane');
 
   // get the lowest directory from the files listing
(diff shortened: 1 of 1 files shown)


guybedford:
@nickL thanks for this PR! Can you clarify here exactly how the wrong directory can be selected? It would really help to understand the fix.

nickL:
@guybedford.  Here's a dummy repo reproducing the problem, https://github.com/nickL/jspm-cli-1737

Creating a bundle with:
 ```
jspm build ./js_apps/folder_with_a_long_name/js_app_1/component_1/application.js  ./js_apps/folder_with_a_long_name/js_app_1/component_1/application.bundle.js  --watch

```
results in a npm module set as the watched folder, in this case `intl`. LMK if that helps.

**Comment by bgoscinski** [comment]
I can reproduce but I have to say that calculating lowestDir basing on path length measured in characters is pretty naive. Consider `./a/b/c/d/e` vs `./a/some_dir1/some_dir2`. I think that what we're trying to achieve here is `lowestDir === './a'` and https://www.npmjs.com/package/commondir seems like does what we want.

**Comment by nickL** [comment]
@bgoscinski: https://www.npmjs.com/package/commondir looks promising, great share!  Yeah testing further char length is limited.  @guybedford I'd be happy to make a few more  changes if you think something like `commondir`  might be a better fit.

**Comment by guybedford** [comment]
Right, so do you think we can just add:

```javascript
if (path.dirname(file).split('/').length < lowestDir.split('/').length)
```
as the fix here?

nickL:
@guybedford: That does it.  Just tested on a few dirs as well. PR  updated. Let me know if you need anything else on this one. HTH

guybedford:
:+1:

PR #1750, opened 2016-04-23
Set relative injected bundle urls

### LINKED ISSUE #1749: Injecting bundles don't work if site root is subfolder
Running `jspm bundle src/main bundle.js --inject` successfully creates a bundle and injects it into my `config.js` file. This all works fine locally but when I push live it breaks because my live site is served from a subfolder and `bundle.js` get's loaded from the root and 404s.

Manually changing `config.js` from something like this:

```js
bundles: {
  "bundle.js": [
    "foo",
    "bar"
  ]
}
```
to this:

```js
bundles: {
  "./bundle.js": [
    "foo",
    "bar"
  ]
}
```

Fixes the issue.

lukechilds:
Fixes #1749 

I'm not sure if there's a better way to do this rather than just concatenating the strings but this is working well for me.

### DIFF
--- a/lib/bundle.js
+++ b/lib/bundle.js
@@ -72,7 +72,7 @@ Builder.prototype.bundle = function(expressionOrTree, outFile, opts) {
     if (opts.injectConfig && opts.outFile) {
       // NB deprecate
       output.bundleName = output.bundleName || self.getCanonicalName(toFileURL(path.resolve(opts.outFile)));
-      var bundleName = output.bundleName;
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks for the PR here.

Unfortunately this approach doesn't work because `./` is taken to be page relative. In an app with pages that have different subpaths this would resolve differently depending on the depth of the subpath (`/page/a -> /page/bundle.js` while `/index -> /bundle.js`).

Issue #1775, opened 2016-05-03
jspm bundle test.js -wid takes 100% of CPU


Vanuan:
I'm following this guide:
http://jspm.io/0.17-beta-guide/development-bundling.html

The problem is that after this line:

```
Watchman:  Watchman was not found in PATH.  See https://facebook.github.io/watchman/docs/install.html for installation instructions
```

jspm is starting to hit 100% of CPU. If I change the file it takes 5 seconds to notice the change and another 10 seconds to actually build it. So those bundles is not a speed up but rather a slowdown of development experience.

Vanuan:
strace shows a lot of such syscalls:

```
inotify_add_watch(11, "/src/node_modules/jspm/node_modules/systemjs-builder/node_modules/babel-core/node_modules/babel-register/node_modules/home-or-tmp/node_modules",
IN_MODIFY|IN_ATTRIB|IN_MOVED_FROM|IN_MOVED_TO|IN_CREATE|IN_DELETE|IN_DELETE_SELF|IN_MOVE_SELF) = 1472
```

Vanuan:
Should it watch for node_modules and jspm_packages files? Is there a way to add them to watch exceptions?

Vanuan:
After several minutes it's dropped to 0% cpu usage. But after each save is goes up again.
It looks like every change creates a new watcher: https://github.com/jspm/jspm-cli/blob/0.17/lib/bundle.js#L279

So there are 2 issues:

1) initial watch is too slow, we should limit a number of files under watch
2) we should reuse watcher and not create a new one on each save

guybedford:
This is a completely valid point, but would involve restructuring to check if the watched module list required for the new bundle matches the previously watched module list of the old bundle, and then only rewatching in the case where it doesn't.

PR work along these lines is incredibly welcome!!

Vanuan:
At least we can skip the `node_modules` folder as you probably won't be installing new modules frequently.
And for now we can just skip the rewatching everything part and address the point "2".

Vanuan:
I've created https://github.com/amasad/sane/issues/66

In either case it seems to me that we should change to glob pattern and use some other directory for auto-generated files.

amasad:
Globs are not meant to be single file names -- this results in o(mn) by just setting up the watch. We end up testing every file in the directory against every file in the glob using the glob library (which expects a glob pattern therefore is more expensive than a string match).

[...]

PR #1776, opened 2016-05-04
Don't create new watcher on each change

Vanuan:
Creating a new watcher is a time consuming operation.
See #1775

### DIFF
--- a/lib/bundle.js
+++ b/lib/bundle.js
@@ -346,11 +346,7 @@ function buildWatch(files, invalidate, build) {
         ui.log('ok', 'File `' + file + '` changed, rebuilding...');
       else
         ui.log('ok', 'File changes made during previous build, rebuilding...');
-      build().then(function(newWatcherBuild) {
-        watcher.close();
-        if (rebuild)
-          newWatcherBuild();
-      });
+      build();
     }
   });
 }
--- a/lib/bundle.js
+++ b/lib/bundle.js
@@ -346,7 +346,9 @@ function buildWatch(files, invalidate, build) {
         ui.log('ok', 'File `' + file + '` changed, rebuilding...');
       else
         ui.log('ok', 'File changes made during previous build, rebuilding...');
-      build();
+      build().then(function() {
+        building = false;
+      });
     }
   });
 }
--- a/lib/bundle.js
+++ b/lib/bundle.js
@@ -348,10 +348,16 @@ function buildWatch(files, invalidate, build) {
       building = true;
       rebuild = false;
       
-      if (filepath) 
-        ui.log('ok', 'File `' + file + '` changed, rebuilding...');
-      else
+      if (filepath) {
+        if(relFiles.indexOf(filepath) == -1) {
+          ui.log('ok', 'File `' + file + '` changed, but not part of the bundle, skipping...');
+        } else {
+          ui.log('ok', 'File `' + file + '` changed, rebuilding...');
+        }
+      }
+      else {
         ui.log('ok', 'File changes made during previous build, rebuilding...');
+      }
       build(file).then(function() {
         building = false;
       });
--- a/lib/bundle.js
+++ b/lib/bundle.js
@@ -175,7 +175,7 @@ exports.bundle = function(moduleExpression, fileName, opts) {
   }
 
   function bundleWatch() {
-    bundleOnly()
+    return bundleOnly()
     .then(function(output) {
       if (!opts.watch)
         return output;
--- a/lib/bundle.js
+++ b/lib/bundle.js
@@ -189,8 +189,14 @@ exports.bundle = function(moduleExpression, fileName, opts) {
         return path.resolve(config.pjson.baseURL, file);
       }), function invalidate(invalidated) {
         systemBuilder.invalidate(toFileURL(invalidated));
-      }, function rebuild() {
-        return bundleOnly();
+      }, function rebuild(triggerFile) {
+        if (triggerFile != fileName &&
+            triggerFile.indexOf('jspm.browser.js') == -1) {
+          // skipping the bundle file
+          return bundleOnly();
+        } else {
+          return Promise.resolve();
+        }
       });
     });
   }
@@ -305,7 +311,7 @@ function buildWatch(files, invalidate, build) {
       return file.substr(lowestDir.length + 1);
     });
 
-    var watcher = sane(lowestDir, { glob: relFiles, watchman: watchman });
+    var watcher = sane(lowestDir, { glob: ['**/*.js', '**/*.json'], watchman: watchman });
     watcher.on('error', function(e) {
       if (e.toString().indexOf('Watchman was not found in PATH') == -1) {
         reject(e);
@@ -346,7 +352,7 @@ function buildWatch(files, invalidate, build) {
         ui.log('ok', 'File `' + file + '` changed, rebuilding...');
       else
         ui.log('ok', 'File changes made during previous build, rebuilding...');
-      build().then(function() {
+      build(file).then(function() {
         building = false;
       });
     }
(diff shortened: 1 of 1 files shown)


Vanuan:
I've just manually tested it and it works. Probably the same should be applied to the build scenario. There's a code duplication to be refactored though.

guybedford:
Are you using the Node native watcher or Watchman?

The reason we rewatch is because one of the invalidated files may have included a new dependency, which would then not be watched with an approach like this I believe?

Vanuan:
I'm using a native watcher.

> The reason we rewatch is because one of the invalidated files may have included a new dependency

So you're saying that we only watch files which are included in the bundle? It doesn't look to be the case.

Vanuan:
We watch every new file in the folders watched:
```
    watcher.on('add', changed);
```

guybedford:
I don't think that ever gets called though because we specifically filter through the glob option to just the known tree of files. That line is kind of there as a precaution but can probably be removed.

Vanuan:
If we filtered it it would't have consumed 100% CPU and wouldn't take so much time

guybedford:
Yes the optimized filtering only applies when using Watchman for a quick refresh, but without Watchman, the globbing should still be applying as we pass it in https://github.com/jspm/jspm-cli/blob/0.17/lib/bundle.js#L305, so I don't think these events will be coming back, but I may be wrong.

Vanuan:
It looks like plain `['**/*.js', '**/*.json']` has a much better performance than hundreds of glob patterns in that array.

Probably, under the hood `sane` does a filesystem scan for each glob pattern or some other inefficient thing.

Vanuan:
But we must limit that glob to only source folder, so that the compiled bundle is out of the watched directory.

Vanuan:
It would be great if we could use this: https://github.com/amasad/sane/pull/50

Vanuan:
Ok, here's the simplest thing that works. But there are some limitations:
1) It works only for 1 bundle. Multiple bundles (if supported) would trigger an infinite loop, so we should have names of all of the bundles for proper filtering
2) This file cries for refactoring. There's a major code duplication for build and bundle functions. Many variables need better names.

guybedford:
Thanks for putting in the effort into this area. It is exactly caring about the code in this sort of way that helps things move forward. In terms of these changes, it is important that code is _correct_ and _maintainable_ is the highest priority before it is _fast_. It is no use having a fast native watcher if the implementation is then not correct. Specifically in this watcher case, it does run with good performance when using Watchman, which is the recommended optimization step currently.

> 1) It works only for 1 bundle. Multiple bundles (if supported) would trigger an infinite loop, so we should have names of all of the bundles for proper filtering

Watched bundles are currently only supported through the CLI, so I don't think this case is possible currently. Definitely worth thinking about in future if we open this up to the API as a watching option though.

> 2) There's *.js extension hard coded, but we'd probably need to watch all the files (*) and have the user decide which files are part of the bundle

This is a no-go unfortunately. We support loading any extension as a module, and plugins can also load any extension.

> 3) If we only watch for files in the bundle, we should parse the jspm.browser.js file on every change of this file and somehow figure out whether it was actually changed or just touched during rebuild (why do we even touch this file on rebuilding?)

Configuration file writes should only happen when it is deemed the configuration file has been altered, it shouldn't be happening everytime, and if it is that is a bug. Configuration changes 

> 4) This file cries for refactoring. There's a major code duplication for build and bundle functions. Many variables need better names.
> 5) Apparently, there are no tests.

Please try to help constructively here. This is 200 lines of code providing a useful first-implementation of a feature. Improvements (along the guidelines provided) are very welcome.

Vanuan:
I'm not sure how to move forward

Vanuan:
The implementation provided in this request looks correct to me.

Vanuan:
As I can see, jspm config isn't part of the bundle currently, so it's not watched. Is it correct?

lib/bundle.js
```diff
@@ -339,14 +348,18 @@ function buildWatch(files, invalidate, build) {
       building = true;
       rebuild = false;
       
-      if (filepath) 
-        ui.log('ok', 'File `' + file + '` changed, rebuilding...');
-      else
+      if (filepath) {
+        if(relFiles.indexOf(filepath) == -1) {
+          ui.log('ok', 'File `' + file + '` changed, but not part of the bundle, skipping...');
```
guybedford: Did you mean to add a skip process here?

lib/bundle.js
```diff
@@ -186,13 +189,19 @@ exports.bundle = function(moduleExpression, fileName, opts) {
         return path.resolve(config.pjson.baseURL, file);
       }), function invalidate(invalidated) {
         systemBuilder.invalidate(toFileURL(invalidated));
-      }, function rebuild() {
-        return bundle();
+      }, function rebuild(triggerFile) {
+        if (triggerFile != fileName &&
+            triggerFile.indexOf('jspm.browser.js') == -1) {
```
guybedford: Won't this filter be already handled by the skipping process below?

lib/bundle.js
```diff
@@ -339,14 +348,18 @@ function buildWatch(files, invalidate, build) {
...
+        } else {
+          ui.log('ok', 'File `' + file + '` changed, rebuilding...');
+        }
+      }
+      else {
         ui.log('ok', 'File changes made during previous build, rebuilding...');
-      build().then(function(newWatcherBuild) {
-        watcher.close();
-        if (rebuild)
-          newWatcherBuild();
```
guybedford: Removing these lines disables the "rebuild" functionality, which is what happens if you edit a file while it is running the watched build, so it knows to immediately rebuild again.

lib/bundle.js
guybedford: Because the rebuild may be triggered by a file change that happened while running a build, we don't necessarily only have one triggerFile. So ideally we can forgo this argument.

lib/bundle.js
```diff
@@ -302,7 +311,7 @@ function buildWatch(files, invalidate, build) {
       return file.substr(lowestDir.length + 1);
     });
 
-    var watcher = sane(lowestDir, { glob: relFiles, watchman: watchman });
+    var watcher = sane(lowestDir, { glob: ['**/*.js', '**/*.json'], watchman: watchman });
```
guybedford: Perhaps just leave out the glob entirely here?

guybedford:
Yes the config file is not watched or checked for changes currently, that would be a feature to add. We can create an issue for this separately if you like.

The implementation seems exactly along the right lines though. Is it providing better performance for you already with this?

lib/bundle.js
Vanuan: It's a debug message to show that only files in the bundle trigger the rebuild.

lib/bundle.js
Vanuan: This is just an example to show the possibilities here. Do you prefer rebuilding on any file change or just the change of file in the bundle? As I can see jspm.browser.js isn't a part of the bundle, so it won't trigger rebuild. We can add it later after we address the bug that leads to this file being touched on rebuild.

lib/bundle.js
Vanuan: Yes, just "*" should be fine.

lib/bundle.js
Vanuan: I'll see what can be done here.

lib/bundle.js
guybedford: I may have misread this, but it looks like it would still be running the build though in this case?

lib/bundle.js
guybedford: Sounds good, keep me posted and let me know if you have any questions about this rebuilding behaviour at all too.

lib/bundle.js
Vanuan: Yeah, the same check should be applied in the `rebuild(triggerFile)`. Or it should be refactored to make it easier to follow.

guybedford:
I went ahead and implemented this watcher sharing in https://github.com/jspm/jspm-cli/commit/8cef864c5b3f95fc41daaae0dc4f725f9204d276.

Issue #1778, opened 2016-05-04
Issue when loading "jspm" in restricted environments and/or with read-only access


swernerx:
We currently experience an issue in our chain of dependencies (postcss-import => pkg-resolve => jspm) where a readonly home directory (or some other write issue with the ~/.jspm folder) lead to a hard exception been thrown by JSPM. 

In this case it is happening for API usage, but it is also valid for the cli tools.

Original issue in "postcss-import" is here:
https://github.com/postcss/postcss-import/issues/198

Output:

```
$ jspm

/usr/local/lib/node_modules/jspm/lib/global-config.js:115
    throw 'Unable to write global configuration file\n' + e.stack;
    ^
Unable to write global configuration file
Error: EACCES: permission denied, open '/Users/Sebastian/.jspm/config'
    at Error (native)
    at Object.fs.openSync (fs.js:584:18)
    at Object.fs.writeFileSync (fs.js:1224:33)
    at save (/usr/local/lib/node_modules/jspm/lib/global-config.js:111:10)
    at Object.<anonymous> (/usr/local/lib/node_modules/jspm/lib/global-config.js:94:3)
    at Module._compile (module.js:413:34)
    at Object.Module._extensions..js (module.js:422:10)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Module.require (module.js:367:17)
```

In this case I made the existing ".jspm" folder read-only. I think that this should never produce a full-blown exception which is preventing the overall functioning of both the API and the basic interaction with the commandline e.g. `jspm --version`.

guybedford:
Released in jspm 0.16.35.

PR #1779, opened 2016-05-04
Fixed importing with non-writeable home- or jspm-folder. Issue #1778.

swernerx:
Instead of failing hard - just through importing - it might be enough to

### DIFF
--- a/lib/global-config.js
+++ b/lib/global-config.js
@@ -90,8 +90,12 @@ dprepend(exports.config, {
   }
 });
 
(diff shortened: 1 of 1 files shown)


swernerx:
This is a fix for #1778 as it does not prevent jspm from loading anymore.

guybedford:
👍

swernerx:
Thanks for the quick response!

Any chance to get this into a release for stable + beta quickly?

PR #1801, opened 2016-05-10
chore(appveyor): test cli on Windows

douglasduteil:
Hi @guybedford 

Going through the [the lastest npm version changelog](https://github.com/npm/npm/releases/tag/v3.9.0), I discovered [AppVeyor](https://ci.appveyor.com).
It's yet another CI but this one is running on Windows :rabbit: 

So I made a [little study](https://github.com/douglasduteil/study-test-es6-universal-lib/pull/1) to propose testing the jspm CLI on a native Windows environnement.

Because Windows users need some love too ;)

PR #1816, opened 2016-05-16
test(appveyor): temporarily allow all job to fail

douglasduteil:
Following #1801

Temporally allow all test to fail on Windows until we resolve them.
Seems like the tests on Windows are failing. So I'm proposing to use them as information to make fixes.
They shouldn't block over PRs.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -6,7 +6,12 @@ environment:
     - nodejs_version: "4"
     - nodejs_version: "6"
 matrix:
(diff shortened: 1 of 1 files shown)


guybedford:
I've finally got the tests passing now on 0.17 and added the bade there.

If we don't add the badge to the master branch, do we still need this?

douglasduteil:
Great :) 
I'm closing this PR then.

Issue #1848, opened 2016-05-25
Bundle build not refreshing / updating html files when changed


abhipanda:
Hi Guy and team,

I am running jspm 0.17.0-beta.15 and using the latest fast build bundle. Running the command

`jspm bundle src/app/app.js app.build.js -wid`

works fine  for JS files but facing issues when changing htmls. Its not updating the module.html tagged with plugin-text

```
 File src/app/modules/my-project/module/module.html changed, rebuilding...
 Building the bundle tree for src/app/app.js...

app/modules/my-project/module/module.html!github:systemjs/plugin-text@0.0.8/text.js
```

Thanks,
Abhi

abhipanda:
Solution:

Change 

`builder.invalidate(toFileURL(file));`

to

`builder.invalidate(toFileURL(file)+'*');`

jonricaurte:
@abhipanda I had to reinstall plugin-jade. Try reinstalling.

guybedford:
Released in 0.17.0-beta.18.

PR #1850, opened 2016-05-26
#1848 - Bundle build not refreshing / updating html files when changed

abhipanda:
Added the wildcard at end of the file URL name to invalidate URL which doesn't match exact URLs


guybedford:
Are you importing `file.html!text` using plugin syntax?

abhipanda:
Yes that and ng-template. 

`import '../shared/templates/some.tpl.html!ng-template';`

guybedford:
Sure, ok I think we should make this:

```javascript
builder.invalidate(toFileURL(file)); 
builder.invalidate(toFileURL(file) + '!*'); 
```

If that doesn't work though just let me know and we should fix the builder itself to allow this.

abhipanda:
Ok sure I will do that

abhipanda:
@guybedford !* won't work for JS files as they don't have an extension.

guybedford:
@abhipanda yes js files will only catch the `builder.invalidate(toFileURL(file));` case, but by doing both we should always catch the right invalidations.

abhipanda:
Ok so you want both to be there. Don't you think thats redundant - I mean first it will try to get through full name and then with wildcard even though it invalidated the first time.

Doing the wildcard will work for both in one execution. Your call.

guybedford:
Yes exactly, invalidating both is the necessary approach despite redundancy.

abhipanda:
Updated !

abhipanda:
hi @guybedford  is this going to be merged soon ?

guybedford:
@abhipanda apologies, I missed your update comment here. Do you think you could rebase the changes to the branch again for merge?

guybedford:
Included manually in https://github.com/jspm/jspm-cli/commit/fd784241e19d85381ffdf61cc7ad2ad4653dcd68. Thanks for your efforts!

abhipanda:
Oh thanks you @guybedford I was going to do it anyways :)

abhipanda:
Hello @guybedford - when is this fix expected into release ?

PR #1870, opened 2016-06-01
[0.17.0-beta-17] Fixed typescipt support for `jspm init` (fixes #1859)

### LINKED ISSUE #1859: Bundling yields error source has multiple anonymous System.register calls.
I switched over from cjs to system and now when I run:
```
jspm bundle src/App.js ./dist/index.js --skip-source-maps
```
I get 
```
err  Source StudioDashboard/App.js has multiple anonymous System.register calls.
```

this is my config:
https://github.com/born2net/studioDashboard/blob/hot-reload/jspm.config.js
and project:
https://github.com/born2net/studioDashboard/tree/hot-reload

any suggestions?

tx

Sean

jakeNiemiec:
- fixed the initialization of typescript projects by loading .ts files into plugin-typescipt instead of *.js files. This eventually resulted in the `err  Source StudioDashboard/App.js has multiple anonymous System.register calls.` error when trying to bundle or build (fixes #1859).

- If the user inputs a typescipt file for the main entry point the 'Select a transpiler' step will default to 'typescript' instead of 'babel'

Note: I tried to match your coding style, making it as terse as possible. Tell me if there is anything wrong with it. Thanks!

### DIFF
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -790,7 +790,7 @@ function initPrompts(newProject) {
     var curTranspiler = loader.transpiler;
 
     // default transpiler will be typescript only if the main loader file ends with .ts
-    var defTranspiler = /(?:\.([^.]+))?$/.exec(loader.package.main)[1] === 'ts' ? 'typescript' : 'babel';
+    var defTranspiler = /\.ts$/.test(loader.package.main) ? 'typescript' : 'babel';
 
     if (curTranspiler && (curTranspiler.substr(0, 7) == 'plugin-' || curTranspiler.substr(0, 7) == 'loader-') && 
         transpilers.indexOf(curTranspiler.substr(7).toLowerCase()) != -1)
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -789,13 +789,16 @@ function initPrompts(newProject) {
 
     var curTranspiler = loader.transpiler;
 
+    // default transpiler will be typescript only if the main loader file ends with .ts
+    var defTranspiler = /(?:\.([^.]+))?$/.exec(loader.package.main)[1] === 'ts' ? 'typescript' : 'babel';
+
     if (curTranspiler && (curTranspiler.substr(0, 7) == 'plugin-' || curTranspiler.substr(0, 7) == 'loader-') && 
         transpilers.indexOf(curTranspiler.substr(7).toLowerCase()) != -1)
       curTranspiler = curTranspiler.substr(7).toLowerCase();
     else if (curTranspiler)
       transpilers.push(curTranspiler);
 
-    return ui.input('%SystemJS.config transpiler (Babel, Traceur, TypeScript, None)%', curTranspiler || 'babel', {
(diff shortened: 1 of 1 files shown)


lib/config/index.js
```diff
@@ -789,13 +789,16 @@ function initPrompts(newProject) {
 
     var curTranspiler = loader.transpiler;
 
+    // default transpiler will be typescript only if the main loader file ends with .ts
+    var defTranspiler = /\.ts$/.test(loader.package.main) ? 'typescript' : 'babel';
```
guybedford: `loader.package` an be null, so perhaps make this `.test(loader.package && loader.package.main)`.

lib/config/index.js
```diff
@@ -812,9 +815,10 @@ function initPrompts(newProject) {
       // set transpiler on BOTH the transpiler and local package loader config
       if (loader.package) {
         var pkgMeta = loader.package.meta = loader.package.meta || {};
-        pkgMeta['*.js'] = pkgMeta['*.js'] || {};
-        if (!pkgMeta['*.js'].loader || pkgMeta['*.js'].loader == loader.transpiler)
-          pkgMeta['*.js'].loader = pkgMeta['*.js'].loader || 'plugin-' + transpiler.toLowerCase();
+        var pkgExt = transpiler === 'typescript' ? '*.ts' : '*.js';
```
guybedford: Sure, makes sense.

guybedford:
So to clarify - the bug here was that the `.js` files were already precompiled by TypeScript so the loader was effectively trying to recompile already-compiled files?

jakeNiemiec:
- While ES5 is normally valid typescript, the output of files transpiled to the system/register module format that get transpiled will throw `error TS2304: Cannot find name 'System'.` Basically:

>`file.ts` --`$ tsc file.ts -m system`-> `file.js` --`$ tsc file.js -m system`-x error TS2304!
File.ts compiles to file.js, which, if file.js is compiled, will throw an error because it `Cannot find name 'System'`.

plugin-typescript expects to be set as a loader for Typescript files (as reflected in the docs). This is a edge case where we are attempting to transpile a module to a module (aka [_I am transpileing a module_](https://en.wikipedia.org/wiki/I_Am_Sitting_in_a_Room)). Furthermore the typescript compiler will reject .js files with `error TS6054: File 'file.js' has unsupported extension. The only supported extensions are '.ts', '.tsx', '.d.ts'.`. It is additionally worth noting that .tsx and .d.ts are "not intended to be implemented by engines or browsers"; therefore I don't see a need to support them within the current scope. 

- The part of this is quality of life improvements for *new* typescript (and potentially new jspm) users.

tl;dr: only `.ts` files should be passed by default to the typescipt transpiler; failing to do so will always(?) cause errors, but may not always reach you. @frankwallis may want to screen incoming files, and error if they are already modules.

guybedford:
Sure, thanks for clarifying, sounds good to me. Just let me know when you've got to the correction above?

jakeNiemiec:
Pushing now, is it customary to squash commits for a PR?

guybedford:
Thanks. Yes it is normally, but I don't mind too much.

Issue #1906, opened 2016-06-12
Non interactive `jspm init` in jspm@0.17.0-beta.20


mrhyde:
when I'm trying to `jspm init` I'm not getting promt to choose options and it fills everything by default.
affected versions are `0.17.0-beta.20` and `0.17.0-beta.19`. `0.17.0-beta.18` works just fine
[screen](http://i.imgur.com/KKAtrGz.png)

dougludlow:
I'm seeing the same thing on beta.20 on OS X. It doesn't wait for prompts, just goes with the defaults.

```bash
$ jspm --version
0.17.0-beta.20
```

dougludlow:
Looks like this line here https://github.com/jspm/jspm-cli/compare/0.17.0-beta.18...0.17.0-beta.20#diff-f265d6a6b6b64d4d096fddcf7d620a76R29 is what introduced the bug. When the `api.js` in imported it sets `ui.useDefaults`. Removing this line https://github.com/jspm/jspm-cli/blob/0.17.0-beta.20/api.js#L51 fixes the issue.

I'm not sure the significance of that line or why it was there historically, @guybedford perhaps you can shed some light on this?

dougludlow:
Perhaps added for tests: https://github.com/jspm/jspm-cli/commit/f91375bf920bb2c0d675c85f4c5c2e02799ed5d2?

dougludlow:
@guybedford I've submitted a pr to remove that line. Tests pass just fine.

guybedford:
Thanks! Fixed in 0.17.0-beta.21.

PR #1907, opened 2016-06-12
[#1906]: Do not set useDefaults to true by default.

dougludlow:


guybedford:
Thanks, I added the fix for this in https://github.com/jspm/jspm-cli/commit/bf31a23ecd2fde660f7bb945ea04803d73bd9192.

PR #1916, opened 2016-06-15
`--watch` deals with the configuration changing.

jackfranklin:
This change means you can install a package and not have your watch task
bail on you.

My thoughts here are to keep an eye out for when the config file changes (when you do an install for example) and then refresh the builder instance such that it loads the new config in.

However doing so in the manner I've got in the code leads to the `bundle` failing with:

```
err  Error: Configuration file jspm.config.js has been modified by another process.
```

Which I completely understand. I'm not sure then if this is straightforward as I think, or if actually I'm just going about it in the whole way - my lack of knowledge of the internals of SystemJS Builder might be holding me back.

@guybedford have you any thoughts?

### DIFF
--- a/lib/bundle.js
+++ b/lib/bundle.js
@@ -150,9 +150,7 @@ exports.bundle = function(moduleExpression, fileName, opts) {
   opts = opts || {};
   fileName = fileName || path.resolve(config.pjson.baseURL, 'build.js');
 
-  function bundle(builder) {
-    var latestBuilder = builder || systemBuilder;
-
+  function bundle(givenBuilder) {
     return Promise.resolve()
     .then(function() {
       if (!opts.sourceMaps)
@@ -161,7 +159,7 @@ exports.bundle = function(moduleExpression, fileName, opts) {
     .then(function() {
       ui.log('info', 'Building the bundle tree for %' + moduleExpression + '%...');
 
-      return latestBuilder.bundle(moduleExpression, fileName, opts);
+      return givenBuilder.bundle(moduleExpression, fileName, opts);
     })
     .then(function(output) {
       logTree(output.modules);
@@ -178,7 +176,7 @@ exports.bundle = function(moduleExpression, fileName, opts) {
     });
   }
 
-  return bundle()
+  return bundle(systemBuilder)
   .then(function(output) {
     if (!opts.watch)
       return output;
@@ -221,9 +219,7 @@ exports.build = function(expression, fileName, opts) {
   opts = opts || {};
   fileName = fileName || path.resolve(config.pjson.baseURL, 'build.js');
 
-  function build(newBuilder) {
-    var latestBuilder = newBuilder || systemBuilder;
-
+  function build(givenBuilder) {
     return Promise.resolve()
     .then(function() {
       if (!opts.sourceMaps)
@@ -232,7 +228,7 @@ exports.build = function(expression, fileName, opts) {
     .then(function() {
       ui.log('info', 'Creating the single-file build for %' + expression + '%...');
 
-      return latestBuilder.buildStatic(expression, fileName, opts);
+      return givenBuilder.buildStatic(expression, fileName, opts);
     })
     .then(function(output) {
       logTree(output.modules, output.inlineMap ? output.inlineMap : opts.rollup);
@@ -253,14 +249,14 @@ exports.build = function(expression, fileName, opts) {
         ui.log('warn', 'Build output to %' + opts.format + '% requires the global name to be set.\n' +
           'Added default %--global-name ' + generatedGlobalName + '% option.\n');
         opts.globalName = generatedGlobalName;
-        return build();
+        return build(systemBuilder);
       }
       else
         throw e;
     });
   }
 
-  return build()
+  return build(systemBuilder)
   .then(function(output) {
     if (!opts.watch)
       return output;
@@ -395,6 +391,7 @@ function buildWatch(builder, output, build) {
       if (changeWasConfigFile) {
         config.loader = new JspmSystemConfig(config.pjson.configFile);
         builder = new Builder();
+        return;
       }
 
       if (file) {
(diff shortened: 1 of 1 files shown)


jackfranklin:
Maybe we need a config method that will just ignore any timestamp conflicts? Otherwise I'm not sure how we can get around this issue.

guybedford:
When the configuration file changes, we just need to trigger the logic to reload the configuration file itself.

This can be done via:

```javascript
config.loader = new JspmSystemConfig(config.pjson.configFile);
```

Just like we do in the main init process - https://github.com/jspm/jspm-cli/blob/0.17/lib/config/index.js#L97.

The `new Builder` should also be done after loading the new config so that the new builder instance will then use that `config.loader` configuration.

guybedford:
Perhaps it may make more sense to encapsulate a `config.reloadConfig()` method that can itself be in the `index.js` file?

guybedford:
Actually I'm not sure... we already have two config load methods there.

jackfranklin:
@guybedford makes sense 👍 

The latest commit adds the line you suggested. We don't now see the error about another process modifying the file, however the build seems to be cached somehow and although it doesn't error it doesn't build with the newest config. You can install jQuery and reference it and the actual build.js file that's produced won't reflect those changes. I have to restart the bundle process for it to pick them up.

I think this could be some logic in my code that's wrong, because I can't see how anything else is cached given that I'm loading in the config & creating the builder instance again from scratch?

guybedford:
That may be due to the `builder` instance not fully propagating when updated in this way? I think there are a few builder closures dotted around?

jackfranklin:
@guybedford I'm an idiot, I was editing `exports.build` but of course I needed to edit `exports.bundle`. I'm tidying up but I think I got it working :)

jackfranklin:
@guybedford can people run `build` with `--watch` ? Should I make the same changes there?

I know this code isn't perfect yet - working on tidying up! Thank you as always for all your help!

guybedford:
Yes build can also be watched. Sure, it looks like the right approach to me. I will review properly when I can as well, just let me know when it's ready.

jackfranklin:
@guybedford this should be in a reviewable state now 👍

jackfranklin:
@guybedford just a little bump in case you'd missed this :)

guybedford:
Haven't missed it - just waiting for the time to properly review!

jackfranklin:
👍 Understood! Sorry to nag prematurely (in hindsight, two days isn't much time!).

lib/bundle.js
```diff
@@ -218,7 +221,9 @@ exports.build = function(expression, fileName, opts) {
   opts = opts || {};
   fileName = fileName || path.resolve(config.pjson.baseURL, 'build.js');
 
-  function build() {
+  function build(newBuilder) {
+    var latestBuilder = newBuilder || systemBuilder;
```
guybedford: It would be nice to keep this function consistent - perhaps let's make `builder` a required argument always for the `build` function, and make sure all calls to `build` provide the builder instance?

lib/bundle.js
```diff
@@ -149,7 +150,9 @@ exports.bundle = function(moduleExpression, fileName, opts) {
   opts = opts || {};
   fileName = fileName || path.resolve(config.pjson.baseURL, 'build.js');
 
-  function bundle() {
+  function bundle(builder) {
+    var latestBuilder = builder || systemBuilder;
```
guybedford: And the same here.

lib/bundle.js
```diff
@@ -391,17 +410,26 @@ function buildWatch(builder, output, build) {
...
-        return buildWatch(builder, output, build);
-      })
-      .then(function(newWatcherRefresh) {
+
+      var buildPromise;
+      if (changeWasConfigFile) {
+        // don't bother doing a rebuild for a config file change
+        // but just keep watching so the next time a file changes
+        // the bundle is rebuilt but with the new Builder & config
+        buildPromise = buildWatch(builder, output, build);
```
guybedford: Is this because there a race condition with injection if we do a rebuild on a config file change?

guybedford: It should be possible to just do a short return on line 398 instead of this approach though.

guybedford:
Thanks for the patience here!

jackfranklin:
@guybedford sorry for a slow reply. I've made both of those changes and things seem to work at my end. Might be worth quickly testing on one of your projects too .

lib/bundle.js
```diff
@@ -248,14 +249,14 @@ exports.build = function(expression, fileName, opts) {
         ui.log('warn', 'Build output to %' + opts.format + '% requires the global name to be set.\n' +
           'Added default %--global-name ' + generatedGlobalName + '% option.\n');
         opts.globalName = generatedGlobalName;
-        return build();
+        return build(systemBuilder);
```
guybedford: givenBuilder?

lib/bundle.js
```diff
@@ -175,7 +176,7 @@ exports.bundle = function(moduleExpression, fileName, opts) {
     });
   }
 
-  return bundle()
+  return bundle(systemBuilder)
```
guybedford: Perhaps make this `return bundle(new Builder())` and remove the `systemBuilder` reference entirely at the top?

lib/bundle.js
```diff
@@ -248,14 +249,14 @@ exports.build = function(expression, fileName, opts) {
...
-        return build();
+        return build(systemBuilder);
       }
       else
         throw e;
     });
   }
 
-  return build()
+  return build(systemBuilder)
```
guybedford: As in line 179.

guybedford:
Sorry to be difficult... it's just quite a complex function (my fault I know!), so trying not to make it more complex...

lib/bundle.js
jackfranklin: This doesn't work because a few lines below there is a reference to `systemBuilder`. What I've done is move the variable down to be with this line so it's clearer. Done this in both places, let me know what you think.

jackfranklin:
@guybedford no need to apologise! Let me know how it looks.

guybedford:
Awesome, thanks Jack for your work here.

jackfranklin:
🎉  thanks for your patience and help!

PR #1968, opened 2016-07-05
inspect: multiple versions detail

oligot:
By running the command `jspm inspect`, we can check if there are
multiple versions of the same dependency in the project.

This commit adds a new flag `--multiple` that allows to have more
details about these multiple versions, notably from which parent
dependency they come from.
This is super useful when trying to understand why there are
multiple versions of the same dependency in the project.

Example output

     Installed Versions

            npm:d3 3.5.17 npm:d3-heatmap@1.2.1
            npm:d3 4.1.0
       npm:isarray 0.0.1 npm:readable-stream@1.1.14
       npm:isarray 1.0.0 npm:buffer@3.6.0

We can see that there are 2 versions of d3:
* 3.5.17, which comes from the parent dependency d3-heatmap
* 4.1.0, which is a direct dependency (as it has no parent)

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -76,7 +76,7 @@ process.on('uncaughtException', function(err) {
       + 'jspm uninstall name                Uninstall a package and clean dependencies\n'
       + 'jspm clean                         Clear unused and orphaned dependencies\n'
       + '\n'
-      + 'jspm inspect [--forks]             View all installed package versions\n'
+      + 'jspm inspect [--forks] [--multiple] View all installed package versions\n'
       + 'jspm inspect npm:source-map        View the versions and ranges of a package\n'
       + '\n'
       + 'jspm inject <name[=target]> [--force] [--latest] [--lock] [-o]\n'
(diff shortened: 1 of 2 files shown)


guybedford:
There's actually a `jspm inspect npm:d3` which shows what loads d3, for exactly this case. Perhaps we can try to make that easier to work with instead of adding it as another feature?

oligot:
Running the command `jspm inspect npm:d3` only gives the installed version of d3, but it doesn't tell us that d3 is also a dependency of d3-heatmap and that they have different versions

    Installed versions of npm:d3
     
            d3 4.1.0 (^4.1.0)

Do you mean we could extend the command `jspm inspect registry:name` to also take into account indirect dependencies instead of adding a new flag to the `jspm inspect` command ?

guybedford:
@oligot `jspm inspect npm:d3` is actually supposed to show all the forks and what their parent packages are, as you describe.

oligot:
Ok, so this seems more like a bug in `jspm inspect registry:name` then.
I'll try to fix it and open a PR when it's done.

PR #1974, opened 2016-07-07
refactor(config): Implement checksum for file integrity

m-a-r-c-e-l-i-n-o:
I'm working on a project that uses programmatic jspm api. To be able to properly unit test, it's necessary to tear down and setup new environments each time, which means multiple `jspm.install(true)` calls. As it stands the following error is getting in the way:
`Configuration file "jspm.config.js" has been modified by another process.`. 
In this pull request, I provide an alternative that serves my needs and continues to serve the intended case (which I assume to be config file integrity and opt-in management). If there is another way to accomplish this without having to modify the code, please do let me know.

### DIFF
--- a/lib/config/config-file.js
+++ b/lib/config/config-file.js
@@ -17,6 +17,7 @@
 var path = require('path');
 var fs = require('fs');
 var mkdirp = require('mkdirp');
+var makeChecksum = require('checksum');
 
 /*
  * Configuration base class
@@ -28,12 +29,12 @@ function ConfigFile(fileName, ordering) {
   this.fileName = path.resolve(fileName);
 
   this.ordering = ordering;
-  
+
   // configuration file style is detected when loading
   this.style = null;
 
-  // we note the configuration file timestamp
-  this.timestamp = null;
+  // we note the configuration file checksum
+  this.checksum = null;
 
   // properties are stored as an ordered array of { key, value } pairs
   // nested objects are in turn array values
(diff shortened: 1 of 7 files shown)


m-a-r-c-e-l-i-n-o:
^^ I doesn't look like that "AppVeyor" build failing do to my changes, let me know if that is not the case. I've verified that all test pass locally as well.

guybedford:
This looks great to me, thanks for posting.

I think it would be better to run a more comprehensive equivalence check here than just looking at the string though.

We have the ability to do this equivalence already by instantiating a new temporary configuration file object, serializing it, and comparing the hash in that serialized space.

Something like - 

```javascript
var md5 = require('../common').md5;

// ...at the end of read, store the full hash...
this.hash = md5(this.serialize());

// to check equivalence with an existing file we can then do the following steps:
var tmpConfigFile = new ConfigFile(this.configFileName);
if (tmpConfigFile.hash != this.hash)
  // -> changed!
```

That way we're running a deeper equivalence, which is much more robust.

Just let me know if that makes sense to update the PR.

guybedford:
Note `new ConfigFile` is not enough - it must be the same class as the parent, so it would need to be `new this.constructor()` I think?

m-a-r-c-e-l-i-n-o:
@guybedford Thanks for the reply, I was under the impression that the originating source was the main priority. I can deff update the pr with the strategy you are proposing (assuming its feasible), but for curiosity sake, what would be scenario where comparing the object (even if it's a deeper check), would yield different results than just comparing the checksum of the originating sources?

guybedford:
@m-a-r-c-e-l-i-n-o this will detect changes like:

```javascript
System.config({
  "a": "b",
  "c": "d"
});
```

as being equivalent to:

```javascript

// whitespace

System.config({
  "a": "b",
  "c": 'd' // one quote not right
});
```

It will still detect style changes though as we do style detection (tabs / spacing / quote style etc).

m-a-r-c-e-l-i-n-o:
@guybedford Got it... I misunderstood your intention. I'll update this pr in the next few days with a solution along the lines of what you are suggesting. Thank you, sir.

guybedford:
@m-a-r-c-e-l-i-n-o that would be amazing, thanks for the contribution.

guybedford:
Closing for lack of activity, but feel free to reopen.

PR #1987, opened 2016-07-18
Fix the jspm inspect registry:name when there are forks

oligot:
Currently, this only shows top level dependencies.
This commit allows to show all the forks and what their parent
packages are.

This was previously discussed in PR #1968 as a new feature,
while it was actually a bug.

Example output:

    Installed versions of npm:d3

       d3 4.1.1 (^4.1.1)

       npm:d3-heatmap@1.2.1
         d3 3.5.17 (^3.5.16)

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -710,7 +710,7 @@ function loadExistingRange(name, parent, inject) {
       });
     })
     .then(function(deps) {
-      return deps[name];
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks so much for looking into this! I've brought it into 0.17 as well.

oligot:
👌

PR #1991, opened 2016-07-20
fix jspm run not exiting with code 1 after error

adamburgess:
similar to my other PRs 🔸 
also happened to once again catch in a CI build.

the direct .catch is for errors in system.import (e.g. file not found, or the module throws on loading) and the uncaughtException exit is for any other error.


guybedford:
Thanks!

Issue #2091, opened 2016-09-21
Add an argument for setting the working directory


jp7677:
Hi there,

Thanks for all you work on jspm. Once you got the hang of it, it is really cool! 
I have a small wish though: A lot of node tools for web development support a working directory as argument, e.g.

```
npm install --prefix ui
ui/node_modules/.bin/typings install --cwd ui
ui/node_modules/.bin/gulp --cwd ui
```
It would be cool if jspm had a similar switch.

guybedford:
Thanks for the feedback, yes something like this would be useful certainly.

jp7677:
Implemented on master and 0.17.

jp7677:
PS: Looking forward to a new release ;).

PR #2114, opened 2016-10-08
Support for setting the working directory.

jp7677:
See https://github.com/jspm/jspm-cli/issues/2091. This pull requests seems to do the magic for me, though I'm sure that it isn't perfect. I'm by no means an experienced javascript developer, so please let me know what I should do better ;).


jp7677:
mmh, don't know why the tests on appveyor fail with my modification, travis seems and local runs (Linux) are fine too.

lib/bundle.js
```diff
@@ -164,7 +165,7 @@ exports.bundle = function(moduleExpression, fileName, opts) {
     if (opts.injectConfig)
       ui.log('ok', '`' + output.bundleName + '` added to config bundles.');
 
-    logBuild(path.relative(process.cwd(), fileName), opts);
+    logBuild(path.relative(env.cwd(), fileName), opts);
```
guybedford: Are you sure this is necessary? This should still work with `process.cwd()` I think?

jspm.js
```diff
@@ -26,7 +27,15 @@ var jspmCLI = new Liftoff({
...
-jspmCLI.launch({}, function(env) {
+
+var args = process.argv.slice(2);
+
+var cwdArgIndex = args.indexOf('--cwd');
+if (cwdArgIndex > -1) {
+  env.setCwd(args[cwdArgIndex + 1]);
+}
+
+jspmCLI.launch({cwd: env.cwd()}, function(env) {
```
guybedford: 👍  just space this - `{ cwd: env.cwd() }`.

Did you test this with both globally and locally installed jspm?

lib/bundle.js
guybedford: It would be nice to do away with the `env` as a separate file and mode here and just work with `process.cwd()` here. Note process.cwd() can be set via https://nodejs.org/api/process.html#process_process_chdir_directory if necessary in the jspm.js bootstrap.

jp7677: Thanks for your feedback, yes using process.chdir(directory) sounds indeed way easier, will do that.

guybedford: Note though that the `cwd` setting in `jspm.js` should be working here for that anyway I think?

jspm.js
jp7677: Only locally ;), thanks for the tip to test globally as well. Will do.

lib/bundle.js
guybedford: Sounds great, just let me know when you've updated the PR here.

guybedford:
Great, thanks for your work here! Would it be possible to port a similar PR for the 0.17 branch as well?

jp7677:
I have updated the PR based on your comments. Seems to work fine locally and globally.

PR #2117, opened 2016-10-10
Improve init defaults for TypeScript projects

### LINKED ISSUE #2110: Improve init defaults for TypeScript projects
TLDR: Improve the getting started experience by

  1. setting "defaultExtenion": "ts" automatically for TypeScript projects, making the jspm-cli even more intuitive.

  2. setting pre-selected transpiler to `typescript` if the local package main ends with `.tsx`

One of the lovely features of jspm is its interactive `init` command. In most cases it sets you up with the right defaults out of the box with an ease I think few cli tools can rival. This includes [default] option selection based on previous entries.
For example, when initializing a new TypeScript project we get
```cmd
~/source/app> jspm init

Init mode (Quick, Standard, Custom) [Quick]: Quick
...
     Enter the main entry point of your package within the src\ folder.

     This allows System.import('app') to resolve to src\/app.js.

SystemJS.config local package main [app.js]:
```
When we enter `app.ts` instead of `app.js`, we are greeted by the pleasant surprise:
```cmd
SystemJS.config local package main [app.js]: app.ts
...
SystemJS.config transpiler (Babel, Traceur, TypeScript, None) [typescript]:
```
`typescript` is pre-selected as the default transpiler, hit enter are we're set.
We get the following automatically
```JavaScript
transpiler: "plugin-typescript",
  packages: {
    "app": {
      "main": "app.ts",
      "meta": {
        "*.ts": {
          "loader": "plugin-typescript"
        }
      }
    }
  }
```
Unfortunately, we will hit errors as soon as we `import` any other .ts files from `app.ts`.
```TypeScript
import {square} from './utils/math'; // src/utils/math.ts
console.log(square(3));
```
// http://localhost:3000/src/utils/math.js 404 (Not Found)
The solution is to add `"defaultExtension": "ts"` to the app package config above.
Note: in versions of TypeScript prior to 2.0.0, it was possible to write.
```TypeScript
import {square} from './utils/math.ts';
```
But this was extremely fragile, destroyed code portability, and was never officially supported by the language.
As of TypeScript 2.0.0 it is a compiler error.

The suggestion here is that we improve the getting started experience by adding `"defaultExtension": "ts"` to the app package automatically for TypeScript projects, making the jspm UX even more intuitive.

A related area of friction is that the following does not work as expected.
```cmd
~/source/app> jspm init

Init mode (Quick, Standard, Custom) [Quick]: Quick
...
SystemJS.config local package main [app.js]: app.tsx
...
SystemJS.config transpiler (Babel, Traceur, TypeScript, None) [babel]:
```
`babel` is pre-selected as the default transpiler, hit enter and we will need to fix the generated config or re-run the command.

The suggestion here is to set the pre-selected transpiler to `typescript` if the local package main ends with `.tsx`. Furthermore, if the above suggestion is adopted it would be natural to set `"defaultExtension": "tsx"`.

I would be happy to submit PRs for any of these changes. 

#2103
#https://github.com/frankwallis/plugin-typescript/issues/162

aluanhaddad:
This resolves #2110 by making the following changes to the `init` command
  1. Include `"defaultExtension": "ts"` in local package config when the `transpiler` is inferred to be TypeScript. For example this will now happen when the local package main is specified as `app.ts`.

  2. Inferring TypeScript as the transpiler when the local package package main has a `.tsx` extension, as should be the case for TypeScript React projects. In this case add `"defaultExtension": "tsx"` to local package config.

### DIFF
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -809,7 +809,7 @@ function initPrompts(newProject) {
       }
     })
     .then(function(transpiler) {
-      if (transpiler === 'none')
+      if (transpiler === curTranspiler || transpiler === 'none')
         return;
 
       // set transpiler on BOTH the transpiler and local package loader config
@@ -827,9 +827,6 @@ function initPrompts(newProject) {
           pkgMeta[pkgExt].loader = pkgMeta[pkgExt].loader || 'plugin-' + transpiler.toLowerCase();
       }
       loader.transpiler = 'plugin-' + transpiler.toLowerCase();
-
-      if (transpiler === curTranspiler)
-        return;
 
       // download transpiler
       var installObj = {};
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -789,8 +789,8 @@ function initPrompts(newProject) {
 
     var curTranspiler = loader.transpiler;
 
-    // default transpiler will be typescript only if the main loader file ends with .ts
-    var defTranspiler = /\.ts$/.test(loader.package && loader.package.main) ? 'typescript' : 'babel';
+    // default transpiler will be typescript only if the main loader file ends with .ts or .tsx
+    var defTranspiler = /\.tsx?$/.test(loader.package && loader.package.main) ? 'typescript' : 'babel';
 
     if (curTranspiler && (curTranspiler.substr(0, 7) == 'plugin-' || curTranspiler.substr(0, 7) == 'loader-') && 
         transpilers.indexOf(curTranspiler.substr(7).toLowerCase()) != -1)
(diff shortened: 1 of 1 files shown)


lib/config/index.js
```diff
@@ -809,19 +809,28 @@ function initPrompts(newProject) {
       }
     })
     .then(function(transpiler) {
-      if (transpiler === curTranspiler || transpiler == 'none')
+      if (transpiler === 'none')
```
guybedford: If I remember correctly, skipping reconfiguration here is important in the case when the transpiler is already configured with manual / custom configuration.

guybedford:
Looks good to me. Is tsx TypeScript + jsx?

lib/config/index.js
aluanhaddad: Gotcha

guybedford:
👍

Issue #1071, opened 2015-08-28
Configure without dependency to github


wgorder:
@guybedford 

I work in at a large enterprise where corporate proxies to github is a problem.  Is there a way to get a successfull JSPM build without any need to talk to github?  I see this as just an example:

```
demo-client  jspm install npm:angular-ui-bootstrap
     Looking up npm:angular-ui-bootstrap
     Updating registry cache...
     Downloading npm:angular-ui-bootstrap@0.13.3
     Looking up npm:angular
     Looking up npm:bootstrap
     Downloading npm:angular@1.4.4

warn Main entry point not found for npm:bootstrap@3.3.5.
     Adjust this property in the package.json or with an override, setting "main": false if this is the intention.

ok   Installed npm:bootstrap@^3.0.0 (3.3.5)
     Looking up github:jspm/nodelibs-path
     Looking up github:jspm/nodelibs-process
     Looking up github:jspm/nodelibs-fs
     Looking up npm:process
ok   Installed github:jspm/nodelibs-path@^0.1.0 (0.1.0)
     Looking up npm:path-browserify
ok   Installed npm:angular (1.4.4)
ok   Installed npm:path-browserify@0.0.0 (0.0.0)
ok   Installed angular-ui-bootstrap as npm:angular-ui-bootstrap@^0.13.3 (0.13.3)
ok   Install tree has no forks.

ok   Install complete.

```

How can I configure things to make the below calls use npm instead?

```
ok   Installed npm:bootstrap@^3.0.0 (3.3.5)
     Looking up github:jspm/nodelibs-path
     Looking up github:jspm/nodelibs-process
     Looking up github:jspm/nodelibs-fs
     Looking up npm:process
ok   Installed github:jspm/nodelibs-path@^0.1.0 (0.1.0)
```

wgorder:
Alternatively if there is a way I can make this work by pointing to an internal Stash repo instead of trying to go to github that would work too.  NPM is mirrored inside via Artrifactory.  If I cant make this work I probably have to rule out JSPM as an option and go another route unfortunately, so I was hoping you could help.

guybedford:
@wgorder is the problem here explicitly the dependency on GitHub, or just getting the proxy to GitHub to behave properly?

Comprehensive proxy support is an issue at the moment I know, and has come up recently a few times.

wgorder:
The problem is a hard dependency on github for our builds to work.  We can work around the proxy  issue in a couple ways but that too has been painful.

[...]

PR #2120, opened 2016-10-11
feat(config): add upgrade code for nodelibs-* deps

frederikschubert:
All github:jspm/nodelibs-* dependencies are being
replaced with npm:jspm-nodelibs-* dependencies.

Resolves https://github.com/jspm/jspm-cli/issues/1071#issuecomment-252886440

### DIFF
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -81,7 +81,7 @@ function updateNodelibTarget(target, range) {
   return target;
 }
 
-function processNodelibs(loader, pjson) {
+function processNodelibs(loader, pjson, upgrade) {
   var installRequired = false;
   Object.keys(loader.baseMap).forEach(function (key) {
     var target = loader.baseMap[key];
@@ -98,12 +98,19 @@ function processNodelibs(loader, pjson) {
     Object.keys(curMap).forEach(function (key) {
       var target = curMap[key];
 
-      if (target.name.substr(0, 21) == 'github:jspm/nodelibs-') {
-        if (key == target.name.substr(21))
-          delete curMap[key];
-        else
+      if (upgrade) {
+        if (target.name.substr(0, 21) == 'github:jspm/nodelibs-') {
           installRequired = true;
           target = updateNodelibTarget(target, false);
+        }
+      } else {
+        if (target.name.substr(0, 21) == 'github:jspm/nodelibs-') {
+          installRequired = true;
+          if (key == target.name.substr(21))
+            delete curMap[key];
+          else
+            target = updateNodelibTarget(target, false);
+        }
       }
     });
   });
@@ -249,7 +256,7 @@ exports.load = function(prompts, allowNoConfig) {
 
       ui.log('info', 'Upgrading jspm 0.16 Node core libraries to jspm 0.17 universal implementations...\n');
 
-      processNodelibs(config.loader, config.pjson);
+      processNodelibs(config.loader, config.pjson, true);
 
       ui.log('info', 'Checking all overrides into the package.json file to ensure reproducibility independent of the jspm registry...\n');
 
@@ -343,7 +350,7 @@ exports.load = function(prompts, allowNoConfig) {
     });
   }))
   .then(function(initInstalls) {
-    var installRequired = processNodelibs(config.loader, config.pjson);
+    var installRequired = processNodelibs(config.loader, config.pjson, false);
     if (installRequired) {
       return require('../install').install(true).then(function () {
         return initInstalls;
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -98,7 +98,7 @@ function processNodelibs(loader, pjson, upgrade) {
     Object.keys(curMap).forEach(function (key) {
       var target = curMap[key];
 
-      if (upgrade) {
+      if (!upgrade) {
         if (target.name.substr(0, 21) == 'github:jspm/nodelibs-') {
           installRequired = true;
           target = updateNodelibTarget(target, false);
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -74,14 +74,10 @@ exports.getLoaderConfig = function() {
   return cfg;
 };
 
-function updateNodelibTarget(target) {
-  target.registry = 'npm';
-  target.name = target.name.replace('github:jspm/nodelibs-', 'npm:jspm-nodelibs-');
-  target.package = target.package.replace('jspm/nodelibs-', 'jspm-nodelibs-');
-  target.exactPackage = target.exactPackage.replace('jspm/nodelibs-', 'jspm-nodelibs-');
-  target.exactName = target.exactName.replace('github:jspm/nodelibs-', 'npm:jspm-nodelibs-');
-  target.exactNameEncoded = target.exactNameEncoded.replace('github:jspm/nodelibs-', 'npm:jspm-nodelibs-');
-  target.setVersion('0.2.0');
+function updateNodelibTarget(target, range) {
+  target.setRegistry('npm');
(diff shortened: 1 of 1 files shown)


frederikschubert:
@guybedford I tried it locally but it did only change _some_ of my `nodelibs-*` dependencies to the npm packages.

lib/config/index.js
```diff
@@ -74,6 +74,71 @@ exports.getLoaderConfig = function() {
...
+    if (target.name.substr(0, 21) == 'github:jspm/nodelibs-') {
+      target = updateNodelibTarget(target);
+      installRequired = true;
+
+      pjson.peerDependencies[key] = target;
+      delete pjson.dependencies[key];
+    }
+  });
+
+  Object.keys(pjson.devDependencies).forEach(function (key) {
```
guybedford: This probably needs peerDependencies as well?

guybedford: (it wasn't necessary before because peerDependencies was only added in 0.17)

lib/config/index.js
```diff
@@ -74,6 +74,71 @@ exports.getLoaderConfig = function() {
...
 
+function updateNodelibTarget(target) {
+  target.registry = 'npm';
+  target.name = target.name.replace('github:jspm/nodelibs-', 'npm:jspm-nodelibs-');
+  target.package = target.package.replace('jspm/nodelibs-', 'jspm-nodelibs-');
+  target.exactPackage = target.exactPackage.replace('jspm/nodelibs-', 'jspm-nodelibs-');
+  target.exactName = target.exactName.replace('github:jspm/nodelibs-', 'npm:jspm-nodelibs-');
+  target.exactNameEncoded = target.exactNameEncoded.replace('github:jspm/nodelibs-', 'npm:jspm-nodelibs-');
+  target.setVersion('0.2.0');
+  return target;
```
guybedford: Rather than this manual modification, we can use `target.setVersion`, `target.setRegistry` and `target.setPackage`.

lib/config/index.js
```diff
@@ -74,6 +74,71 @@ exports.getLoaderConfig = function() {
...
+    var curMap = loader.depMap[key];
+
+    Object.keys(curMap).forEach(function (key) {
+      var target = curMap[key];
+
+      if (target.name.substr(0, 21) == 'github:jspm/nodelibs-') {
+        if (key == target.name.substr(21))
+          delete curMap[key];
+        else
+          installRequired = true;
```
guybedford: This should probably just be at the top of the entire block?

lib/config/index.js
```diff
@@ -74,6 +74,71 @@ exports.getLoaderConfig = function() {
   return cfg;
 };
 
+function updateNodelibTarget(target) {
```
guybedford: There should probably be a second argument here - `range` as boolean. We can then use `setVersion('^0.2.0)` for the package.json dependencies to ensure we get semver upgrade paths. While the depMap set will be range: false.

lib/config/index.js
frederikschubert: Done.

lib/config/index.js
frederikschubert: I am not exactly sure what you mean here.

lib/config/index.js
frederikschubert: Done.

lib/config/index.js
guybedford: The original version of this code was designed to have two separate branches:

* If the submap name matches the node core lib name (`fs: github:jspm-nodelibs-fs@...`) then it deletes the key entirely. This is for the beta upgrade path so we turn Node core dependencies from normal dependencies into peer dependencies.
* If the submap name doesn't match (the else above in the original code) then update the target version (`custom-fs-name: github:nodelibs-fs@...`). That's a very rare case for jspm-only workflows but what this was designed for in the original code to handle this upgrade comprehensively.

In the adjusted code above `installRequired` is now set on the else path only, and `target` is always updated on all paths due to the lack of curly braces on the conditional.

So there are two problems here:

1. The code above is malformed, at least from a tabbing perspective.
2. We probably do need another flag to indicate the difference between the beta upgrade path and the nodelibs upgrade path. In the first case we want the original behaviour to be maintained. In the second case we just need to do replacement (the target update line below), and probably don't even need the conditional above.

Let me know if that makes sense.

guybedford:
Thanks for the updates here. Were you able to test out the upgrade path again?

frederikschubert:
Not yet. But will try to test it later today!

lib/config/index.js
frederikschubert: Alright. I wasn't aware of this distinction.
How do you propose to detect the difference between the beta upgrade path and the nodelibs upgrade path? I am not familiar enough with the code base to come up with something reasonable.

lib/config/index.js
```diff
@@ -74,6 +74,79 @@ exports.getLoaderConfig = function() {
   return cfg;
 };
 
+function updateNodelibTarget(target, range) {
```
guybedford: function (target, range, upgrade)

lib/config/index.js
```diff
@@ -176,49 +249,7 @@ exports.load = function(prompts, allowNoConfig) {
...
-        var target = config.pjson.devDependencies[key];
-
-        if (target.name.substr(0, 21) == 'github:jspm/nodelibs-') {
-          target.setVersion('^0.2.0-alpha');
-
-          config.pjson.peerDependencies[key] = target;
-          delete config.pjson.devDependencies[key];
-        }
-      });
+      processNodelibs(config.loader, config.pjson);
```
guybedford: processNodelibs(config.loader, config.pjson, true)

lib/config/index.js
```diff
@@ -312,6 +343,16 @@ exports.load = function(prompts, allowNoConfig) {
     });
   }))
   .then(function(initInstalls) {
+    var installRequired = processNodelibs(config.loader, config.pjson);
```
guybedford: processNodelibs(config.loader, config.pjson, false)

lib/config/index.js
frederikschubert: I guess you meant
```ts
function processNodelibs(loader, pjson, upgrade)
```
?

lib/config/index.js
```diff
@@ -74,6 +74,86 @@ exports.getLoaderConfig = function() {
...
+    }
+  });
+
+  Object.keys(loader.depMap).forEach(function (key) {
+    var curMap = loader.depMap[key];
+
+    Object.keys(curMap).forEach(function (key) {
+      var target = curMap[key];
+
+      if (upgrade) {
```
guybedford: I think this should be `if (!upgrade)`.

guybedford:
Yeah that's it, just check the one comment there.

frederikschubert:
I just tested the changes but it does not seem to work yet.

guybedford:
Ok, I can look at this, but it will just take me a while to find the time.

guybedford:
👍 

Merged manually, and made some minor corrections in https://github.com/jspm/jspm-cli/commit/ffadde59a4c7a789576c49cab92698f866accf3b. The main issue I think was also bumping the npm registry cache version to ensure we don't use previously cached dependency values (https://github.com/jspm/npm/commit/7e03fea971dcf66db4dc370bb2d82ef2291a87a5).

frederikschubert:
Great! 👍

guybedford:
Another issue here was we missed the transitive deps - https://github.com/jspm/nodelibs-path/commit/366c5b05e5ee6feed820ab674207a4643290fe37. I think that's all the cases now.

Issue #1899, opened 2016-06-10
Install hooks


guybedford:
An idea for supporting hooks:

``` json
{
  "jspm": {
    "hooks": {
      "preinstall": "./hooks/preinstall.js"
    }
  }
}
```

Where we load the module `./hooks/preinstall.js` and run the default export function with some arguments to run the hook.

Hooks would only be supported in the package doing the install - hooks in installed packages would not run for security reasons.

The set of hooks and arguments should be determined here.

guybedford:
Note that above effectively allows an ecosystem of hooks via:

```
jspm install custom-jspm-hook --dev
```

```json
{
  "jspm": {
    "hooks": {
      "preinstall": "custom-jspm-hook"
    }
  }
}
```

frederikschubert:
This proposal would fit our use case quiet well!

aluanhaddad:
Is this still being considered? I have a node cli tool which attempts to install specifically versioned TypeScript declaration files based on the state of `jspm.config.js` after an install and I would like to make it part of my build process.

WorldMaker:
I have similar but subtly different use case as well: some of the packages I install were built with Typescript themselves and their npm packages and/or github projects include a `tsconfig.json` file. Visual Studio tries to recurse the project folder for all `tsconfig.json` files and build them, resulting in build errors from the `jspm_packages` directly even if I trying to tell VS to ignore that directory. It would be nice if jspm didn't install those files at all (some sort of install file ignore) or I could hook a cleanup process post-install.

aluanhaddad:
@WorldMaker There should be a way to exclude these files in your root tsconfig.json. If not I would file an issue at [Microsoft/TypeScript](//github.com/Microsoft/TypeScript/). The information in these tsconfig.json files is often helpful and I would hate to see them ignored by default.

WorldMaker:
@aluanhaddad They are definitely excluded in the top level tsconfig and Typescript has no issue with it; it just seems to be VS trying to be more helpful than is actually helpful.

aluanhaddad:
What version of VS and what version of TypeScript Tools are you using?

WorldMaker:
[...]

PR #2121, opened 2016-10-12
feat(hooks): add postinstall and preinstall hook support

frederikschubert:
Part of https://github.com/jspm/jspm-cli/issues/1899

I went with the simplest solution of just executing a hook script with no arguments.
@guybedford Could you have a look at this and tell me, if it is alright. If so I will add a `preinstall` hook [here](https://github.com/jspm/jspm-cli/blob/master/lib/install.js#L88).

### DIFF
--- a/lib/config/package.js
+++ b/lib/config/package.js
@@ -37,6 +37,7 @@ var config = require('./index');
  * - peerDependencies
  * - devDependencies
  * - overrides
+ * - hooks
  *
  * Public Methods
  * - write
@@ -85,9 +86,11 @@ function PackageConfig(fileName) {
       'dependencies',
       'devDependencies',
       'peerDependencies',
-      'overrides'
+      'overrides',
+      'hooks'
     ]],
-    'overrides'
+    'overrides',
+    'hooks'
   ]);
 
   this.dir = path.dirname(path.resolve(fileName));
@@ -106,6 +109,7 @@ function PackageConfig(fileName) {
     this.jspmPrefix = true;
 
   this.name = prefixedGetValue.call(this, ['name'], 'string') || !this.jspmAware && 'app';
+  this.hooks = prefixedGetObject.call(this, ['hooks'], true);
 
   var baseURLValue = prefixedGetValue.call(this, ['directories', 'baseURL'], 'string') || '';
   if (baseURLValue[0] == '/' || baseURLValue.indexOf('//') != -1 || baseURLValue.indexOf('\\\\') != -1 || baseURLValue.indexOf(':') != -1) {
--- a/lib/config/package.js
+++ b/lib/config/package.js
@@ -87,10 +87,16 @@ function PackageConfig(fileName) {
       'devDependencies',
       'peerDependencies',
       'overrides',
-      'hooks'
+      ['hooks', [
+        'preinstall',
+        'postinstall'
+      ]]
     ]],
     'overrides',
-    'hooks'
+    ['hooks', [
+        'preinstall',
+        'postinstall'
+    ]]
   ]);
 
   this.dir = path.dirname(path.resolve(fileName));
--- a/lib/install.js
+++ b/lib/install.js
@@ -86,6 +86,24 @@ exports.install = function(targets, options) {
   options = options || {};
 
   return config.load()
+  .then(function () {
+    var hooks = config.pjson.hooks;
+    if (hooks) {
+      return new Loader().import(hooks.preinstall)
+        .then(function (m) {
+          if (m.default && typeof m.default === 'function') {
+            ui.log('info', 'Running preinstall hook...');
+            return Promise.resolve(m.default());
+          }
+        })
+        .catch(function (error) {
+          if (error) {
+            ui.log('err', 'Error during preinstall hook');
+            ui.log('err', error);
+          }
+        });
+    }
+  })
   .then(function() {
     installed = installed || config.loader;
     secondaryRanges = secondaryRanges || config.deps;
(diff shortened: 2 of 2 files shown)


fracz:
+1

Izzmo:
Would love to see this integrated in.

lib/install.js
```diff
@@ -142,6 +143,15 @@ exports.install = function(targets, options) {
       return install(name, targets[name], opts);
     }))
     .then(function() {
+      var hooks = config.pjson.hooks;
+      if(hooks) {
+        if('postinstall' in hooks && typeof hooks.postinstall === 'string') {
+          System.config(config.getLoaderConfig());
+          return System.import(hooks.postinstall);
```
guybedford: Use `api.import` here rather - https://github.com/jspm/jspm-cli/blob/master/api.js#L73.

lib/install.js
```diff
@@ -142,6 +143,15 @@ exports.install = function(targets, options) {
...
+        }
+      }
+    })
+    .then(function() {
       return saveInstall();
```
guybedford: Are you sure we don't want to run the hook after `saveInstall` so that the new configuration files have been saved already?

lib/install.js
guybedford: The only problem with a direct import approach is that it won't work so well for hooks that need to delay further operations as we only get a synchronous execution context (until top-level await becomes a thing that is!).

If this is fine for you I'm happy to go with this, but it could be worth thinking about.

lib/install.js
```diff
@@ -144,6 +147,15 @@ exports.install = function(targets, options) {
     .then(function() {
       return saveInstall();
     })
+    .then(function () {
+      var hooks = config.pjson.hooks;
+      if (hooks) {
+        if ('postinstall' in hooks && typeof hooks.postinstall === 'string') {
+          System.config(config.getLoaderConfig());
```
guybedford: This is no longer necessary.

lib/install.js
```diff
@@ -26,6 +26,7 @@ var PackageName = require('./package-name');
 var ui = require('./ui');
 var path = require('path');
 var globalConfig = require('./config/global-config');
+var System = require('systemjs');
```
guybedford: This can also be removed.

lib/install.js
```diff
@@ -144,6 +147,15 @@ exports.install = function(targets, options) {
     .then(function() {
       return saveInstall();
     })
+    .then(function () {
+      var hooks = config.pjson.hooks;
+      if (hooks) {
+        if ('postinstall' in hooks && typeof hooks.postinstall === 'string') {
+          System.config(config.getLoaderConfig());
+          return api.import(hooks.postinstall);
```
guybedford: If we use this pattern we will cache the loader configuration which may cause issues with ensuring we use the latest configurations.

It may be better to actually use `var Loader = api.Loader` and then `return new Loader().import(hooks.postinstall)`.

Also we should include a catch handler that gives some error information about the hook failure.

guybedford:
Great, any feedback on the fact that the hook is only running as a sync execution context?

frederikschubert:
The only thing that is happening during an install, on which a postinstall hook would depend on, would be the updated configuration files.
I have no use case in mind where the execution would have to be delayed.

guybedford:
Consider though a preinstall hook. For example that may want to do some work asynchronously before the install starts.

frederikschubert:
So do you suggest to rather call the hook's default export function with a promise, that the postinstall hook has to resolve after it is done?

guybedford:
Yes exactly, if it exists and is a function.

frederikschubert:
Alright. I will give it a try and will also add the preinstall hook in the same way.

frederikschubert:
Thanks a lot for your guidance!

lib/config/package.js
```diff
@@ -85,9 +86,11 @@ function PackageConfig(fileName) {
       'dependencies',
       'devDependencies',
       'peerDependencies',
-      'overrides'
+      'overrides',
+      'hooks'
     ]],
-    'overrides'
+    'overrides',
+    'hooks'
```
guybedford: We can actually describe the shape of the hooks object here:

```javascript
['hooks': [
  'preinstall',
  'postinstall'
]]
```

lib/install.js
```diff
@@ -84,6 +86,24 @@ exports.install = function(targets, options) {
   options = options || {};
 
   return config.load()
+  .then(function () {
+    var hooks = config.pjson.hooks;
+    if (hooks) {
```
guybedford: `if (hooks && hooks.preinstall)` surely?

lib/install.js
```diff
@@ -84,6 +86,24 @@ exports.install = function(targets, options) {
...
+      return new Loader().import(hooks.preinstall)
+        .then(function (m) {
+          if (m.default && typeof m.default === 'function') {
+            ui.log('info', 'Running preinstall hook...');
+            return Promise.resolve(m.default());
+          }
+        })
+        .catch(function (error) {
+          if (error) {
```
guybedford: Even if the error is undefined we'd still want the message to display. So perhaps just:

```javascript
if (error)
  ui.log('err', error);
```

lib/install.js
```diff
@@ -144,6 +164,24 @@ exports.install = function(targets, options) {
...
+              return Promise.resolve(m.default());
+            }
+          })
+          .catch(function (error) {
+            if (error) {
+              ui.log('err', 'Error during postinstall hook');
+              ui.log('err', error);
+            }
+          });
+      }
```
guybedford: Since this code is a direct duplication of the above, it is probably worth generalising it into a function.

Happy to have that function at the top of this `install.js` file for the mean time - it doesn't yet have to be its own file.

lib/install.js
frederikschubert: Done.

lib/config/package.js
frederikschubert: Done.

lib/install.js
frederikschubert: Done.

guybedford:
👍

trusktr:
Awesome. Is there a usage example?

frederikschubert:
@trusktr You can add hooks to the project's `package.json` like this.
```json
"jspm": {
    "hooks": {
        "postinstall": "some-module"
    }
},
...
```
`some-module` can be any module that jspm can load. So it could be a file in your project or a dependency that was installed via jspm. The hook has to have the following form.
```js
export default function examplePostInstallHook() {
    // Code
}
```
If the hook does something async, the promise from that operation has to be returned.

Issue #1977, opened 2016-07-11
install latest version of dependency as an exact dependency


joeldenning:
http://jspm.io/docs/installing-packages.html indicates that you can install the exact version of a dependency by doing `jspm install npm:react@15.2.1`, but I don't see a way to install exact the latest version of a dependency without specifying the version you want.

With npm, I do this with `npm install react@latest --save --save-exact` or just `npm install react --save --save-exact` or just `npm i react -S -E` and it automatically puts the latest version as an exact dependency. Would perhaps a `--save-exact` or a `--exact` flag be the right thing for jspm?

guybedford:
The way to handle this use case currently is that it does just mean making it a top-level dependency.

joeldenning:
How does making it a top-level dependency solve this? Doing `jspm install npm:react` will put `"react": "npm:react@^15.3.0"` into my package.json, and I'm looking for a way to make it put in `npm:react@15.3.0` (pinned without the `^`). Am I missing something?

guybedford:
I'm just stating the obvious - that you need to install the sub-dependency first to an exact range in order to handle this use case today.

guybedford:
Released in 0.17.0-beta.32.

PR #2137, opened 2016-10-22
Implement #1977: --exact flag for pinning latest version of dependency

joeldenning:
See #1977. The `--exact` flag is intended to be similar to npm's [save-exact](https://docs.npmjs.com/cli/install) flag, where you don't have to specify the version but it still pins the dependency to the latest version.

I saw that you put the discussion label on #1977 and am not sure if that means there was more to talk about before creating a pull request -- if so let me know.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -70,6 +70,7 @@ process.on('uncaughtException', function(err) {
       + '  install react --lock           Stable install, locking existing dependencies\n'
       + '  install react=npm:react --edge Install a package alias to latest unstable\n'
       + '  install react --quick          Quick install, skipping tree hash verifications\n'
+      + '  install react --exact          Install and pin the latest version of a package\n'
       + '\n'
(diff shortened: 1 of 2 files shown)


guybedford:
This looks great, thanks. Yes it would be definitely worth supporting this.

I'm just wondering if it might be nice to also have this work with target versions:

```
jspm install  react@0.14 --exact
```

Where the above will resolve the range, but still pin the exact version into the package.json.

I think that might be something like this path: (although it can probably be better refactored into the conditional)

```javascript
if (!options.parent && options.exact) {
  target.setVersion(resolution.version);
}
```

FDiskas:
:non-potable_water:

joeldenning:
Sounds good -- I have updated the pull request to support `--exact` with target versions (e.g., `jspm install  react@0.14 --exact`)

guybedford:
👍  thanks for your work here!

PR #2162, opened 2016-11-07
update typescript option to install latest stable

### LINKED ISSUE #2145: trying to use typescript 2
Installing typescript 2 ...  **jspm install npm:typescript** ,  getting this message:

**warn typescript@2.0.6 is unsupported for this version of jspm. Use jspm dl-loader --latest to update.**

Then exec **jspm dl-loader --latest** ... and so the typescript is downgraded to 1.6.2 ... 

It is possible to use typescript 2 with jspm ?

Thanks!

aluanhaddad:
The code emit changes between these major versions all expand TypeScript's support for transpiling ES2015 syntax. A few involve better SystemJS support and I think this upgrade will be beneficial overall. That said, I primarily use TypeScript via plugin-typescript.

This would resolve #2145

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -45,7 +45,7 @@ var tPackages = {
   'core-js':         'npm:core-js@^1.1.4',
   'traceur':         'github:jmcriffey/bower-traceur@0.0.93',
   'traceur-runtime': 'github:jmcriffey/bower-traceur-runtime@0.0.93',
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks! Yes in theory this is a breaking change, but we at least give users the ability to opt-into selecting their own TypeScript version anyway too.

PR #2172, opened 2016-11-15
only rewrite package.json if it has changed

tamird:
@guybedford

### DIFF
--- a/lib/config/package.js
+++ b/lib/config/package.js
@@ -160,10 +160,9 @@ PackageJSON.prototype.read = function(prompts, sync) {
 };
 
 PackageJSON.prototype.write = function() {
-  var self = this;
   var pjson = this.jspmPrefix ? {} : this.originalPjson;
 
-  if (!this.jspmPrefix && self.registry !== 'jspm')
+  if (!this.jspmPrefix && this.registry !== 'jspm')
     pjson.jspm = true;
 
   // set the registry property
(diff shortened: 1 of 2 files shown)


lib/config/package.js
```diff
@@ -227,8 +227,15 @@ PackageJSON.prototype.write = function() {
     this.originalPjson.jspm = pjson;
   }
 
-  // NB check that the file hasn't changed since we opened it and if so, prompt
-  return asp(fs.writeFile)(this.fileName, stringify(this.originalPjson) + config.newLine);
+  asp(fs.readFile)(self.fileName).then(function(content) {
+    if (content === stringify(this.originalPjson) + config.newLine) {
```
guybedford: `this` must be `self`.

guybedford: This may also need to be `content.toString()` to do a correct equality.

lib/config/package.js
```diff
@@ -227,8 +227,15 @@ PackageJSON.prototype.write = function() {
...
+      return Promise.resolve();
+    } else {
+      return asp(fs.writeFile)(this.fileName, stringify(this.originalPjson) + config.newLine);
+    }
+  }).catch(function() {
+    return asp(fs.writeFile)(this.fileName, stringify(this.originalPjson) + config.newLine);
```
guybedford: What is this case for? We should specific the file code error specifically that you are trying to protect against here (if any?).

tamird:
@guybedford updated, PTAL.

guybedford:
:+1:

tamird:
Any chance of getting this released?

PR #2182, opened 2016-11-21
fix setting of jspm and deps prefix

sidloki:
`depsPrefixed` is not set correctly, if no `package.json` exists.

Set `depsBase` has no effect because the value is going to be overridden based on the `depsPrefixed` value when writing the file.

### DIFF
--- a/lib/config/package.js
+++ b/lib/config/package.js
@@ -212,17 +212,13 @@ PackageConfig.prototype.setPrefix = function(jspmPrefix) {
 
     this.file.remove(['jspm']);
 
-    this.depsPrefixed = false;
     this.file.changed = true;
   }
   else if (!this.jspmPrefix && jspmPrefix) {
-    this.jspmPrefix = true;
-    this.depsBase = ['jspm'];
-
     if (this.file.getValue(['jspm']))
       this.file.remove(['jspm']);
   }
-  this.jspmPrefix = jspmPrefix;
+  this.jspmPrefix = this.depsPrefixed = jspmPrefix;
 };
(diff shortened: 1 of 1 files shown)


lib/config/package.js
```diff
@@ -212,17 +212,13 @@ PackageConfig.prototype.setPrefix = function(jspmPrefix) {
 
     this.file.remove(['jspm']);
 
-    this.depsPrefixed = false;
     this.file.changed = true;
   }
   else if (!this.jspmPrefix && jspmPrefix) {
-    this.jspmPrefix = true;
-    this.depsBase = ['jspm'];
```
guybedford: Isn't this line still needed?

guybedford:
Thanks, just let me know about that above line removal?

lib/config/package.js
sidloki: No, the assignments are unneeded:

- `this.jspmPrefix`  is set [here](https://github.com/sidloki/jspm-cli/blob/a58db636e0a96027c9af0aca229a9ed4016d23db/lib/config/package.js#L221) in the same function
- `this.depsBase` is not used anywhere. In the write and init function `depsBase` is generate from `this.depsPrefix` (see [here](https://github.com/sidloki/jspm-cli/blob/a58db636e0a96027c9af0aca229a9ed4016d23db/lib/config/package.js#L262) and [here](https://github.com/sidloki/jspm-cli/blob/a58db636e0a96027c9af0aca229a9ed4016d23db/lib/config/package.js#L142))

Issue #2197, opened 2016-12-14
[0.17] Dependency missing in depcache despite having been installed by jspm itself


Rush:
I'm running 0.17.0-beta.32.  The gyst of the problem is that jspm does not "see" installed dependencies of package `angular-ui-router` as shown on this generated depcache. Trying to run the app with this depcache will yield `angular-ui-router` being loaded without `angular` and breaking.

```
    "depCache": {
      "test.js": [
        "angular-ui-router"
      ]
```

Please see list of commands 

```
rush@rushpod: ~/c/jspm-bug 318ms » jspm init

warn Running jspm globally, it is advisable to locally install jspm via npm install jspm --save-dev.

Package.json file does not exist, create it? [Yes]: Yes
Init mode (Quick, Standard, Custom) [Quick]: Quick
Local package name (recommended, optional): app
package.json directories.baseURL: .
package.json configFiles folder [./]: ./
Use package.json configFiles.jspm:dev? [No]: No
SystemJS.config browser baseURL (optional): /
SystemJS.config Node local project path [src/]: src/
SystemJS.config local package main [app.js]: app.js
SystemJS.config transpiler (Babel, Traceur, TypeScript, None) [babel]: babel
     Updating registry cache...

ok   Installed dev dependency plugin-babel to npm:systemjs-plugin-babel@^0.0.17 (0.0.17)
     Install tree has no forks.
     
ok   jspm init completed. For the standard project layout see http://jspm.io/0.17-beta-guide/creating-a-project.html.
     To rerun these init prompts at any time use jspm init.
     
ok   package.json at package.json
     Config at jspm.config.js
     Looking up loader files...
       system.js
       system.js.map
       system-polyfills.js
       system-polyfills.js.map
       system.src.js
       system-polyfills.src.js
     
     Using loader versions:
       systemjs@0.19.41
ok   Loader files downloaded successfully
rush@rushpod: ~/c/jspm-bug 18.77s » 
rush@rushpod: ~/c/jspm-bug » jspm install angular-ui-router

warn Running jspm globally, it is advisable to locally install jspm via npm install jspm --save-dev.
     Updating registry cache...
     Downloading github:angular-ui/angular-ui-router-bower@0.3.2

warn Unable to use the GitHub API to speed up dependency downloads due to lack of authorization
     To resolve use jspm registry config github to configure the credentials, or update them in your ~/.netrc file.
ok   Installed github:angular/bower-angular (1.6.0)
[...]

PR #2200, opened 2016-12-15
Fixing package overrides where shim name ends with '.js'. See #2197

joeldenning:
See #2197. The package override in the jspm registry for angular-ui-router@0.3.1 does not work anymore because `upgradePackageConfig` appends an extra `.js` to `shim` names that already have a `.js` in them. See [example of a broken package override](https://github.com/jspm/registry/blob/master/package-overrides/github/angular-ui/angular-ui-router-bower%400.3.1.json), [the jspm commit that caused the bug](https://github.com/jspm/jspm-cli/commit/3944cb3bf346a0f47b107937e2de5683f96c944b#diff-8eaf344b0d37588331d22edb952ddd05R320), and [the issue reported for angular-ui-router](https://github.com/jspm/jspm-cli/issues/2197)

### DIFF
--- a/lib/package.js
+++ b/lib/package.js
@@ -306,7 +306,8 @@ function upgradePackageConfig(packageConfig) {
         continue;
 
       var curShim = packageConfig.shim[s];
-      var curMeta = packageConfig.meta[s + '.js'] = {};
+      var metaName = s.endsWith('.js') ? s : s + '.js';
+      var curMeta = packageConfig.meta[metaName] = {};
 
       if (curShim instanceof Array) {
         curMeta.deps = curShim;


joeldenning:
For angular-ui-router@0.3.1, the bug is manifested with the following override being added to the package.json:

```json
    "overrides": {
      "github:angular-ui/angular-ui-router-bower@0.3.1": {
        "dependencies": {
          "angular": "*"
        },
        "meta": {
          "release/angular-ui-router.js.js": {
            "deps": [
              "angular"
            ],
            "format": "global"
          }
        }
      },
      "github:angular/bower-angular@1.6.0": {
        "main": "angular",
        "meta": {
          "angular.js": {
            "exports": "angular",
            "format": "global"
          }
        }
      }
    }
```

guybedford:
@joeldenning shim names are supposed to be without the ".js" as shim was added back when `defaultJSExtensions` was the default setting. So we should rather create overrides that set shim without the extension.

Issue #2174, opened 2016-11-16
JSPM Build Error when minifying in ESM format


leepowellcouk:
When attempting to run a jspm build and output as a ES6 module format I get an error at the uglify stage. So this:

    jspm build app/index.js dist/app.js --minify --skip-source-maps --format esm

Returns 
   
    err  Error: SyntaxError: Unexpected token: keyword (default) (line: 9403, col: 7, pos: 312642)
    Error
        at new JS_Parse_Error (eval at <anonymous> (/Users/leepowell/Sites/app/node_modules/uglify-js/tools/node.js:28:1), <anonymous>:1545:18)
    ...

However, when run without the `---minify` flag it compiles successfully.

The app at this stage is VERY simple with only one or two ES6 modules being loaded in.

guybedford:
Can you share the unminified output here?

leepowellcouk:
Looks like it's the sfx-esm builder template?
[https://gist.github.com/leepowellcouk/f78ba8e38a016f9f941c26ba755ee4d3](https://gist.github.com/leepowellcouk/f78ba8e38a016f9f941c26ba755ee4d3)

guybedford:
Ah right. Yeah the issue here is that UglifyJS doesn't support ES module syntax at all.  So you'd want to output UMD or something else rather to enable minification.

To support ES6 minification we can upgrade to using Babili. There's an issue tracking this in https://github.com/systemjs/builder/issues/726, and if anyone is interested in working on this please let me know.

aluanhaddad:
I think a lot of folks will run into this issue. I wonder if we should emit an error indicating that the options are currently incompatible. We would need to remove it when https://github.com/systemjs/builder/issues/726 is resolved.

Something like
```bash
jspm build app --minify --format esm

err  Error: Option --minify is incompatible with option --format esm, consider specifying a 
different format or skipping minification
```

PR #2215, opened 2017-01-17
build: Explain incompatibility between --format esm and --minify

aluanhaddad:
Based on https://github.com/jspm/jspm-cli/issues/2174 and on my own experience running into the same incompatibility, I think it would be beneficial to provide a warning when combining the `--format esm` and `--minify` options to the build command until the SystemJS Builder is enhanced to use Babili for minification (https://github.com/systemjs/builder/issues/726).


guybedford:
Sure, thanks.

PR #2277, opened 2017-04-12
trying to fix flickering peer dependency range issue

### LINKED ISSUE #2027: jspm install changing peer dependency version of react when running on a different machine.
I have this situation. My existing `package.json` & `jspm.config.js` have react peer dependency  as `"npm:react@^15.3.0"` but running `jspm install` in a different machine changes react version to `0.14.8`

I do recall that on the machine where I first did the set up, it asked me to select  react version, and I chose `0.15` ? So shouldn't it be fixed to `0.15` where ever I run `jspm install`

What could be causing this ?

zenorbi:
Hopefully fixes #2027, would need further edge case investigation.


guybedford:
Thanks @zenorbi.

In your comment you said it fixed the `aurelia-bootstrapper` case. Can you confirm if it completely solved it there for you? I'll test this out as soon as I can.

zenorbi:
Looks like the initial install (`jspm install aurelia-bootstrapper -f --yes`) results in flickering `aurelia-templating`. ^1.3.0 in one run, ^1.0.0 in another, but when ^1.0.0 is written, my code never sees the ^1.3.0 semver. Multiple `jspm install` results in it being ^1.3.0 without flickering. Does the codepath differ between `jspm install` and `jspm install package-name` in some way, maybe prefiltered?

guybedford:
@zenorbi would this fix need to be made in the peerDependency resolution process itself rather at `resolvePeerConflicts` to be fully deterministic / converging?

zenorbi:
Going to take a look at this tomorrow, resolvePeerConflicts looks to be a good place, however I'm struggling to create a consistent repro, maybe as a custom registry and faking latency.

zenorbi:
I added a repro where the issue comes up with a README on how to use the repro script.

zenorbi:
I really have no idea how to fix this. `resolvePeerConflicts` has too little information on this. It only gets an accepted range and one version proposal. I was trying to fix it after `resolvePeerConflicts` where I have two ranges to choose from, but that part of the code doesn't write anything to pjson.

Where can I choose between two ranges (resolved version satisfies both ranges) and write the stricter one to pjson?

guybedford:
Ok I get you. I did some hacking on `resolvePeerConflicts` to see if we could potentially add the additional argument to it, but for now since your approach works, I think I'd rather merge and release that.

Thanks so much for the clear replication case as well, it helped to get a picture for this.

Let's clean this up and merge then if you're still keen? Otherwise I can do this manually.

guybedford:
If we can make the reproduction a completely isolated test case that could be an option, but otherwise we should remove it.

That makes sense. We don't do write-based diffing though - the `storeResolution` method is exactly the layer to do this as the representation of the package.json is what is changed there (think of this like a write to the package.json, just that the actual fs happens at the end). So I think the method we have here sounds right.

zenorbi:
Moved flickering-peer-repro to its own repository. I tried making an isolated test case, but that would require accessing private parts of the api and resetting the internal state after the test.

jpray:
I know 0.17 is more or less abandoned but I'm still wondering would it would take to get this merged and released.  Apart from this issue, 0.17 has been pretty solid for me and this fix would let me stay on it until jspm 2 is released.  I'm happy to pitch in if there is something specific to be done.  Thanks.

guybedford:
Actually I'm just going to go ahead and merge this. If there are issues we can always revert.

PR #2291, opened 2017-05-12
Update global-config.js

SerkanSipahi:
show warn if jspm config is used and no <options>, <setting> were passed

### DIFF
--- a/lib/global-config.js
+++ b/lib/global-config.js
@@ -127,7 +127,7 @@ exports.save = save;
 exports.set = function(name, val) {
  
   if(!(name && val)) {
-     throw 'Please pass option and setting value! See for more information jspm --help';
+     console.warn('Please pass option and setting value! See for more information jspm --help');
   }
   
   var nameParts = name.split('.');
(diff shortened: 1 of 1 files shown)


SerkanSipahi:
@guybedford do you know why is appveyor is failing?

lib/global-config.js
```diff
@@ -125,6 +125,11 @@ function save() {
 exports.save = save;
 
 exports.set = function(name, val) {
+ 
+  if(!(name && val)) {
+     console.warn('Please pass option and setting value! See for more information jspm --help');
+  }
```
guybedford: This check should be in https://github.com/jspm/jspm-cli/blob/0.17/cli.js#L590 I think rather, where we could perhaps include something like:

```
ui.warn('jspm config requires a property and value via %jspm config property value%')
```

or similar.

guybedford:
Thanks for the PR, this seems useful!

The Appveyor build is a little unstable at times unfortunately.

SerkanSipahi:
@guybedford 
>The Appveyor build is a little unstable at times unfortunately.
I would be better if we remove "Appveyor" for a while as long as "Appveyor" is not stable!

PR #2322, opened 2017-06-28
Added warning that module names exist in URL space

CxRes:
Added a warning that module names exist in URL space, so should be / separated. In reference to systemjs/builder#814.

### DIFF
--- a/docs/production-workflows.md
+++ b/docs/production-workflows.md
@@ -6,11 +6,12 @@ There are three main workflows for production:
 
 ### Creating a Bundle
 
+***Important**: The module names such as `app/main` used in the examples below should only be `/` separated (On Windows, do NOT use `\` as your path separator for this argument). The module names are specified in URL space; in particular, they are not file-paths.*
+
 ```
   jspm bundle app/main build.js
 ```
-
-Creates a file `build.js` containing `app/main` and all its dependencies referenced in config.js.
+Creates a file `build.js` containing `app/main` and all its dependencies referenced in config.js. \
 
 We can then load this with a script tag in the page:
 


docs/production-workflows.md
```diff
@@ -6,11 +6,12 @@ There are three main workflows for production:
...
 ### Creating a Bundle
 
+***Important**: The module names such as `app/main` used in the examples below should only be `/` separated (On Windows, do NOT use `\` as your path separator for this argument). The module names are specified in URL space; in particular, they are not file-paths.*
+
 ```
   jspm bundle app/main build.js
 ```
-
-Creates a file `build.js` containing `app/main` and all its dependencies referenced in config.js.
+Creates a file `build.js` containing `app/main` and all its dependencies referenced in config.js. \
```
aluanhaddad: Why the trailing `/`?

CxRes: To force a newline! I guess what happens is that in my editor, md renders differently from github (despite being set for gfm). It should be trivial enough for you to fix :)

docs/production-workflows.md
```diff
@@ -6,11 +6,12 @@ There are three main workflows for production:
 
 ### Creating a Bundle
 
+***Important**: The module names such as `app/main` used in the examples below should only be `/` separated (On Windows, do NOT use `\` as your path separator for this argument). The module names are specified in URL space; in particular, they are not file-paths.*
```
aluanhaddad: I think this should be reworded slightly. It should say something closer to what the title of this PR says. Also, I think **Note**: would be better than **Important**.

CxRes:
I don't mind rewording as long as users are not confused about it (like I
was)...

aluanhaddad:
Agreed.

Issue #2333, opened 2017-07-20
jspm install npm:app-decorators => err  Repo npm:babel-types  not found!


SerkanSipahi:
#### jspm 0.17.0-beta.42

**Details:**
when i run `jspm install npm:app-decorators` i get following error:
```
     Updating registry cache...
ok   Installed npm:babel-preset-app-decorators@0.8.228 (0.8.228)
ok   Installed npm:babel-preset-es2015@6.24.1 (6.24.1)
ok   Installed npm:babel-plugin-transform-class-properties@6.24.1 (6.24.1)
ok   Installed npm:babel-plugin-app-decorators-component@0.8.228 (0.8.228)
ok   Installed npm:babel-plugin-app-decorators-component-register@0.8.228 (0.8.228)
ok   Installed npm:babel-plugin-transform-decorators-legacy@1.3.4 (1.3.4)
ok   Installed npm:babel-plugin-app-decorators-view-precompile@0.8.228 (0.8.228)
ok   Installed npm:babel-plugin-check-es2015-constants@^6.22.0 (6.22.0)
ok   Installed npm:babel-plugin-transform-function-bind@6.22.0 (6.22.0)
ok   Installed npm:babel-plugin-transform-es2015-arrow-functions@^6.22.0 (6.22.0)
ok   Installed npm:babel-plugin-transform-es2015-block-scoping@^6.24.1 (6.24.1)
ok   Installed npm:babel-plugin-transform-es2015-block-scoped-functions@^6.22.0 (6.22.0)
ok   Installed npm:babel-plugin-transform-es2015-modules-amd@^6.24.1 (6.24.1)
ok   Installed npm:babel-plugin-transform-es2015-computed-properties@^6.24.1 (6.24.1)
ok   Installed npm:babel-plugin-transform-es2015-literals@^6.22.0 (6.22.0)
ok   Installed npm:babel-plugin-transform-es2015-duplicate-keys@^6.24.1 (6.24.1)
ok   Installed npm:babel-plugin-transform-es2015-modules-commonjs@^6.24.1 (6.24.1)

err  Repo npm:babel-types  not found!
```
but when i run just `npm:babel-types` everything is fine:
```
     Updating registry cache...
ok   Installed npm:to-fast-properties@^1.0.1 (1.0.3)
ok   Installed npm:esutils@^2.0.2 (2.0.2)
ok   Installed npm:regenerator-runtime@^0.10.0 (0.10.5)
[...]

PR #2340, opened 2017-07-21
backtickets are not shown when using jspm

SerkanSipahi:
backtickets are not shown when using jspm on the command line with nodejs see #2333


ArmorDarks:
Well, the backtickets are kind-of standard for displaying code, snippets, routes etc...

SerkanSipahi:
@ArmorDarks but it will not displayed when it using with throw('.....`....`....'); on command line!

PR #2356, opened 2017-08-17
Update to systemjs 0.20.17

### LINKED ISSUE #2354: New dev-release with newest SystemJS?
When can we expect a new release of the 0.17 branch with the newest SystemJS release?
SystemJS 0.20.15 fixed a circular linking race condition, that's a big issue here.
We can of course run `jspm dl-loader --latest` but it will be much easier for us (and over CI) to just install JSPM without any more stuff to do

aluanhaddad:
Fixes #2354


guybedford:
I'll do this for 0.20.18 just now.

PR #2358, opened 2017-08-20
fix some issues with the --inspect command

### LINKED ISSUE #2231: Equivalent of `npm ls` for JSPM-cli ?
Hi !

I was wondering if JSPM-cli expose a way to do the equivalent of `npm ls name[@version]`.  
The  goal is to show a list of all the versions installed for a package, and a list of every places in the dependency tree where the package is used, for every version.  

I often found that when I want to do `jspm resolve --only registry:name@version`, I don't actually know every places each version of a package is used.  

The only thing I found was which dependency it affected afterwards.

aluanhaddad:
fix some issues with the --inspect command
 - resolves #2231
 - fix spurious "not found"
 - show specified dependency relative to parents

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -859,23 +859,29 @@ function visitForkRanges(tree, resolution, name, parentName, visit) {
   Object.keys(tree.depMap).forEach(function(parent) {
     var curDepMap = tree.depMap[parent];
 
-    Object.keys(curDepMap).forEach(function(dep) {
-      var secondary = curDepMap[dep];
+    Object.keys(curDepMap)
+      .filter(function(dep) {
+        return matchesSecondaryDepName(dep) && isNotForkOfSelf(dep) && hasRange(dep);
+      })
+      .forEach(function(dep) {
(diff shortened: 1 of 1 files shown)


lib/install.js
```diff
@@ -859,23 +859,29 @@ function visitForkRanges(tree, resolution, name, parentName, visit) {
...
-      // skip if we don't have a range
+    function matchesSecondaryDepName(dep) {
+      return curDepMap[dep].name === resolution.name;
+    }
+    function isNotForkOfSelf(dep) {
+      return dep !== name || parent !== parentName;
+    }
+    function hasRange(dep) {
       var ranges = secondaryRanges[parent];
-      if (!ranges || !ranges[dep])
```
guybedford: Would the simple fix not be to just adjust this to `!ranges.deps[dep]`?

I know your adjustment is easier to read, but involves unnecessary repeated dictionary lookups, so in the name of performance I think it may be best to keep this unrolled.

guybedford:
Thanks so much for tracking this down finally!

PR #2378, opened 2017-10-18
fix: showInstallGraph fails if a config is missing from package.json

aluanhaddad:
When running
```shell
$jspm inspect registry:depname
```
where depname is mapped in a configuration file but is not a dependency in `package.json`, an error is thrown by the `showInstallGraph` method. This happens because the `range` argument is `undefined` under these circumstances.

This happens after the warning is correctly printed, but is distracting, making the warning itself easy to miss.

This fix falls back to using the value of `resolved.version` which is likely not ideal behavior either.
Perhaps we should print `'unspecified'` or something of the sort.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -1017,12 +1017,19 @@ function showInstallGraph(pkg) {
     ui.log('info', '\nInstalled versions of %' + pkg.name + '%');
     visitForkRanges(installed, pkg, null, null, function(name, parent, resolved, range) {
       found = true;
-      if (range.version === '')
-        range.version = '*';
-      var rangeVersion = range.name === resolved.name ? range.version : range.exactName;
-      if (range.version === '*')
-        range.version = '';
-
+      var rangeVersion;
+      if (range) {
+        if (range.version === '') {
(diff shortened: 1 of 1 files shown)


lib/install.js
```diff
@@ -1017,12 +1017,19 @@ function showInstallGraph(pkg) {
...
+        rangeVersion = range.name === resolved.name ? range.version : range.exactName;
+        if (range.version === '*') {
+          range.version = '';
+        }
+      }
+      else {
+        rangeVersion = resolved.version; 
+      }
       if (!parent)
         ui.log('info', '\n       %' + name + '% `' + resolved.version + '` (' + rangeVersion + ')');
```
guybedford: Perhaps just handle `rangeVersion` as `undefined` here meaning the brackets are just skipped or replaced with `(unspecified)` as you say.

guybedford:
Thanks, will be great to fix this.

lib/install.js
aluanhaddad: Roger

PR #2380, opened 2017-10-25
update to the latest systemjs-plugin-babel

aluanhaddad:
This fixes a chalk issue, described in https://github.com/systemjs/plugin-babel/issues/80, that causes failures in Windows environments.

This has been causing the AppVeyor CI builds to fail consistently for a while now.


guybedford:
Thanks for tracking this down!

PR #2381, opened 2017-10-27
Bump jspm-github

Y--:


aluanhaddad:
@guybedford it looks like the Travis CI builds are failing for the reasons discussed in https://github.com/jspm/registry/issues/1064 and resolved in https://github.com/jspm/registry/pull/1065

Master currently builds against 0.10, 0.11, and 0.12.
If we need to support node < 4, we may need to lock down our dependency on `request` which depends on `hawk` and it in turn on `hoek`.

guybedford:
@aluanhaddad thanks, I think it may well be worth updating CI to Node 4 at this point.

aluanhaddad:
Created #2382

PR #2459, opened 2019-05-13
Fix packages folder path in mapping

jbanety:
My `jspm_packages` folder could be in a subfolder with in this config :
```json 
"directories": {
  "baseURL": "public"
}
```

### DIFF
--- a/src/map/index.ts
+++ b/src/map/index.ts
@@ -181,7 +181,7 @@ class Mapper {
       return await cached;
     
     return await (this.cachedPackagePaths[pkgName] = (async () => {
-      const pjson = await readJSON(`${this.project.projectPath}/jspm_packages/${pkgName.replace(':', '/')}/package.json`);
+      const pjson = await readJSON(`${this.project.config.pjson.packages}/${pkgName.replace(':', '/')}/package.json`);
 
       if (!pjson)
         throw new JspmUserError(`Package ${highlight(pkgName)} is not installed correctly. Run jspm install.`);
--- a/src/map/index.ts
+++ b/src/map/index.ts
@@ -287,7 +287,7 @@ class MapResolver {
   mapResolve: (id: string, parentUrl: string) => string;
 
   constructor (project: Project, map: ImportMap) {
-    let baseDir = project.projectPath;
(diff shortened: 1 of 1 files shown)


jbanety:
With these fixes, it raised an error : 
```bash
$ jspm trace --deps ./public/src/app.js
```
```
err  TypeError: Cannot read property 'registry' of undefined
    at conditionMap ([...]/node_modules/@jspm/resolve/resolve.js:1175:14)
    at applyMap ([...]/node_modules/@jspm/resolve/resolve.js:1204:16)
    at [...]/node_modules/jspm/lib/map/index.js:100:39
```

guybedford:
I was typing my reply to note about the issue as you hit it :)

In previous versions we used to refer to this as `"directories": {  "packages" }`.

I went back and fourth on implementing this in jspm 2.0 because for example `npm` doesn't let you control that `node_modules` should be in a different folder to the package.json and yet no one seems to complain about that.

Also there are some complexities here in that the location of jspm_packages affects the resolver quite deeply. I did support this in the resolver an earlier iteration of jspm 2.0 in https://github.com/jspm/jspm-resolve/blob/1e2ab08eeb9ed55424b5df5523ccb8bfc32f30bc/resolver-spec.md#project-configuration, which could be brought back too.

I'm open to working on this, but let's perhaps discuss why this is different from npm a little first too.

guybedford:
I do prefer your referring to it as `baseURL` over it being the `packages` configuration though actually.

guybedford:
Btw if you want to join jspm chat on Discord, you can follow this link - https://discord.gg/fkGfpH.

jbanety:
My main goal is to serve static files from a `/public` folder with Express.
So I want to put `jspm_packages` in a subfolder.

guybedford:
Let's keep this open as a feature discussion.

It's possible but it adds complexity to the resolver - so I specifically only want to implement it if it is really deemed important.

guybedford:
Ahh, wait this is a PR. I've created a tracking issue in https://github.com/jspm/jspm-cli/issues/2464.

PR #2469, opened 2019-05-16
Add no-relative-deps option to trace cmd

jbanety:
Some of my module import module with relative path e.g. `import aaa from "./dir/aaa.js"`.
When I run the command `jspm trace --deps ./app.js` it outputs : 
```
@scope/package
@otherscope/package2
./dir/aaa.js
```
wich is correct. 

But if I want to use the commands you noted in your guide at section `Optimized Dependency Builds` :
```
$ jspm build $(jspm trace --deps ./app.js) -h -o deps-buildmap.json
$ jspm build ./test.js --external deps-buildmap.json
```
it will fail with : 
```
$ jspm build $(jspm trace --deps ./app.js) -h -o deps-buildmap.json
err  Error: Could not load [projectBasePath]/dir/aaa.js: ENOENT: no such file or directory, open '[projectBasePath]/dir/aaa.js'
```

So my proposal is to add an option to exclude relative dependencies as they will be bundle with the second build.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -277,7 +277,7 @@ ${bold('Command Flags')}
       case 't':
       case 'trace': {
         let options;
-        ({ args, options } = readOptions(args, ['react-native', 'production', 'electron', 'node', 'deps'], ['out', 'in']));
+        ({ args, options } = readOptions(args, ['react-native', 'production', 'electron', 'node', 'deps', 'no-relative-deps'], ['out', 'in']));
 
         project = new api.Project(projectPath, { offline, preferOffline, userInput, cli: true });
         const map = await api.map(project, options);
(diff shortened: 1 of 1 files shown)


guybedford:
I believe the intent of the `--deps` flag was that it should only list the "plain resolution" dependencies in the first place.

Could you update this PR to make this the default behaviour for `--deps` instead of being a new flag?

jbanety:
Is `dep.startsWith(".") === false` the right way to check ? It looks too simple :-)

guybedford:
The definition of bare specifiers is that they are not URLs (in theory this
means failing URL parse), and they do not start with ‘./‘ or ‘../‘.

There’s a utility function for the URL check in
https://github.com/jspm/jspm-cli/blob/master/src/utils/common.ts#L86.

jbanety:
I created an other PR #2472.

PR #2472, opened 2019-05-17
Make sure --deps only list the "plain resolution" dependencies

jbanety:


guybedford:
Are you sure that is sufficient? I think it still needs the
startsWith(‘./‘) || startsWith(‘../‘) check?

jbanety:
Yes. I added it

guybedford:
Released in 2.0.0-beta.4.

PR #2476, opened 2019-05-20
Override pkg config with jspm section if it exists

### LINKED ISSUE #2467: jspm section in package.json
Hi @guybedford,
I created several jspm packages that are used as dependencies for my main application.
With jspm 0.17 there was a jspm section in the package.json file.
With 2.0, when I install the jspm packages, the dependencies in the jspm section are not installed with.
Do I have to put these dependencies at the root of the package.json file? But I do not want them to be installed with `yarn install` because this packages are only for the front.

Thanks.

jbanety:
Go with https://github.com/jspm/jspm-resolve/pull/9 and fix #2467

### DIFF
--- a/src/install/package.ts
+++ b/src/install/package.ts
@@ -420,6 +420,12 @@ export function validateOverride (pcfg: PackageConfig, name: string) {
 }
 
 export function processPackageConfig (pcfg: PackageConfig, partial = false, registry = ''): ProcessedPackageConfig {
+
+  // Override with jspm section if it exists.
+  if (typeof pcfg.jspm === 'object') {
+    Object.assign(pcfg, pcfg.jspm);
+  }
+
   const processed: ProcessedPackageConfig = processPjsonConfig(pcfg);
   if (processed.peerDependencies)
     delete processed.peerDependencies;


src/install/package.ts
```diff
@@ -420,6 +420,12 @@ export function validateOverride (pcfg: PackageConfig, name: string) {
 }
 
 export function processPackageConfig (pcfg: PackageConfig, partial = false, registry = ''): ProcessedPackageConfig {
+
+  // Override with jspm section if it exists.
+  if (typeof pcfg.jspm === 'object') {
```
guybedford: When overriding dependencies it is important that any `jspm.dependencies` means all dependency properties are overridden.

This can be captured with the following logic:

```js
if (pcfg.jspm && (pcfg.jspm.dependencies || pcfg.jspm.devDependencies || pcfg.jspm.peerDependencies || pcfg.jspm.optionalDependencies)) {
  pcfg.dependencies = pcfg.jspm.dependencies || {};
  pcfg.devDependencies = pcfg.jspm.devDependencies || {};
  pcfg.peerDependencies = pcfg.jspm.peerDependencies || {};
  pcfg.optionalDependencies = pcfg.jspm.optionalDependencies || {};
}
```

Happy to merge with that change.

PR #2480, opened 2019-06-23
Allow version tags in aliases

Hypercubed:
Here is PR for allowing version tags in aliases for your review.

Note: I had a hard time compiling this for testing.  I was seeing many TS errors that I needed to hack around.  What is your build process?  Can we add that to the package scripts?


Hypercubed:
Here are my TS fixes:  https://github.com/Hypercubed/jspm-cli/tree/fix/ts-issues

Not sure if you need these.

src/cli.ts
```diff
@@ -31,7 +31,7 @@ import publish from './install/publish';
 import { getBin } from './install/bin';
 import { spawn } from 'child_process';
 
-const installEqualRegEx = /^(@?([-_\.a-z\d]+\/)?[\-\_\.a-z\d]+)=/i;
+const installEqualRegEx = /^(@?([-_\.a-z\d]+\/)?[\-\_\.a-z\d]+(@[-_\.a-z\d]+)?)=/i;
```
guybedford: Perhaps we should just simplify this entirely to just `/^[-_\.a-z\d@]+=/i`?

src/utils/common.ts
```diff
@@ -272,6 +272,7 @@ export function md5 (input: string): string {
 }
 
 export const validPkgNameRegEx = /^(@[-_\.a-z\d]+\/)?[-_\.a-z\d]+$/i;
+export const validAlaisRegEx = /^(@[-_\.a-z\d]+\/)?[-_\.a-z\d]+(@[-_\.a-z\d]+)?$/i;
```
guybedford: See suggested simplification above. Also note `validAlaisRegEx` -> `validAliasRegEx`.

src/install/index.ts
```diff
@@ -417,16 +417,18 @@ export class Installer {
       // auto-generate name from target if necessary
       if (!install.name) {
         if (typeof install.target !== 'string') {
-          if (install.target.name.substr(install.target.name.lastIndexOf(':') + 1).match(validPkgNameRegEx))
-            install.name = install.target.name.substr(install.target.name.lastIndexOf(':') + 1);
+          const idx = install.target.name.lastIndexOf(':') + 1;
+          const substr = install.target.name.substr(idx);
+          if (substr.match(validAlaisRegEx))
```
guybedford: I'm not sure I follow this part since `install.target` is supposed to be the `registry:name@version` thing right? So in `jspm install @x=x`, the `install.target` should be the `npm:x` part right? Not the `@x` part? It's been a while since I've looked at this code, so maybe you can clarify this for me?

src/cli.ts
Hypercubed: I tried to keep it as strict as it was before.  If `/^[-_\.a-z\d@]+=/i` looks good to you I can make that change and update the PR.

Hypercubed:
@guybedford Did you see the changes here: https://github.com/Hypercubed/jspm-cli/tree/fix/ts-issues?  Not sure if or how this was working previously.... but I need that to run the tests.

src/install/index.ts
Hypercubed: This was not really necessary for this PR.  This was a refactor to reduce some redundant code while I was debugging.

Hypercubed:
I updated this PR including the changes I need to get TS to run.

guybedford:
Published in 2.0.0-beta.5. Thanks so much for the ts build fixes too. Usually we don't commit the build changes, and only do that for release, but it was fine here.

jbanety:
Hi,
This PR introduces some bugs.
Command :   
`jspm build $(jspm trace --deps ./public/src/app.js) -h --dir ./public/dist -o ./public/dist/deps-buildmap.json`
   
Error raised :
```
err (jspm) TypeError: Function has non-object prototype 'undefined' in instanceof check   
    at Function.[Symbol.hasInstance] (<anonymous>)    
    at isRecord (/jspm/lib/build/index.js:40:18)   
    at Object.build (/jspm/lib/build/index.js:43:9)   
    at cliHandler (/jspm/lib/cli.js:602:44)   
    at Object.<anonymous> (/jspm/lib/cli.js:699:5)   
    at Module._compile (internal/modules/cjs/loader.js:776:30)   
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:787:10)   
    at Module.load (internal/modules/cjs/loader.js:653:32)   
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)     
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
```

guybedford:
@jbanety thanks for reporting - I believe this should be fixed in https://github.com/jspm/jspm-cli/pull/2482.

guybedford:
Released in 2.0.0-beta.6.

Hypercubed:
Oof...  that was a half completed change.  Thanks!

PR #2484, opened 2019-06-30
Fix windows loader path

### LINKED ISSUE #2483: Cannot find module C:\Program
I run jspm following the docs, installing http-server
```
jspm install http-server --dev
jspm_packages/.bin/http-server.cmd
```

In the result I see:

```
internal/modules/esm/default_resolve.js:59
  let url = moduleWrapResolve(specifier, parentURL);
            ^

Error: Cannot find module 'C:\Program' imported from C:\projects\spect\
    at Loader.resolve [as _resolve] (internal/modules/esm/default_resolve.js:59:13)
    at Loader.resolve (internal/modules/esm/loader.js:70:33)
    at Loader.getModuleJob (internal/modules/esm/loader.js:143:40)
    at Loader.import (internal/modules/esm/loader.js:127:28)
    at internal/process/esm_loader.js:44:43
    at Object.exports.initializeLoader (internal/process/esm_loader.js:51:5)
    at initializeESMLoader (internal/bootstrap/pre_execution.js:373:9)
    at prepareMainThreadExecution (internal/bootstrap/pre_execution.js:52:3)
    at internal/main/run_main_module.js:7:1 {
  code: 'ERR_MODULE_NOT_FOUND'
```

Seems that it doesn't handle spaces in path properly

dy:
Fixes #2483 for windows.

### DIFF
--- a/src/install/bin.ts
+++ b/src/install/bin.ts
@@ -63,7 +63,7 @@ const winBin = (binModulePath: string) => `@setlocal
   @echo jspm not found in path, make sure it is installed globally.
   @exit /B 1
 )
-@set NODE_OPTIONS=--experimental-modules --no-warnings --loader /%JSPM_PATH%node_modules\\jspm\\node_modules\\@jspm\\resolve\\loader.mjs
+@set NODE_OPTIONS=--experimental-modules --no-warnings --loader "%JSPM_PATH%node_modules\\jspm\\node_modules\\@jspm\\resolve\\loader.mjs"
 @node "%~dp0\\..\\${binModulePath}" %*
 `;
 


src/install/bin.ts
```diff
@@ -63,7 +63,7 @@ const winBin = (binModulePath: string) => `@setlocal
   @echo jspm not found in path, make sure it is installed globally.
   @exit /B 1
 )
-@set NODE_OPTIONS=--experimental-modules --no-warnings --loader /%JSPM_PATH%node_modules\\jspm\\node_modules\\@jspm\\resolve\\loader.mjs
+@set NODE_OPTIONS=--experimental-modules --no-warnings --loader "%JSPM_PATH%node_modules\\jspm\\node_modules\\@jspm\\resolve\\loader.mjs"
```
guybedford: This should still be `"/%JSPM_PATH` to ensure correct normalization cases of the URL.

guybedford (approved):
Thanks for the PR, will merge and release asap.

src/install/bin.ts
dy: Hm. This gives `Error: Cannot find module 'C:\C:Program Filesnodejsnode_modulesjspmnode_modules@jspmresolveloader.mjs' imported from C:\projects\spect\`.

In fact, in some strange way the loader defined this way seems to be ignored by node. That's weird, because putting `loader` file in relative path as `./loader.mjs` works.

guybedford: That's really odd.... so the point here is that `C:\\Program Files...` is not a valid path for the loader normalization because it will do `fileURLToPath(new URL('C:\\Program Files', pathToFileURL(process.cwd())))` effectively. And `C:` is not the protocol, rather `file:` must be the protcol, so we get that by adding the leading `/`.

What is the `%JSPM_PATH%` here for you? I'm still not sure I follow how it is working without the leading `/`.

dy: `@echo %JSPM_PATH%` gives `C:\Program Files\nodejs\`. Hmm.

dy: And the latter still doesn't work - it gives
```
internal/modules/esm/default_resolve.js:59
  let url = moduleWrapResolve(specifier, parentURL);
            ^

Error: Cannot find module '' imported from C:\projects\spect\
    at Loader.resolve [as _resolve] (internal/modules/esm/default_resolve.js:59:13)
    at Loader.resolve (internal/modules/esm/loader.js:70:33)
    at Loader.getModuleJob (internal/modules/esm/loader.js:143:40)
    at Loader.import (internal/modules/esm/loader.js:127:28)
    at internal/process/esm_loader.js:44:43
    at Object.exports.initializeLoader (internal/process/esm_loader.js:51:5)
    at initializeESMLoader (internal/bootstrap/pre_execution.js:373:9)
    at prepareMainThreadExecution (internal/bootstrap/pre_execution.js:52:3)
    at internal/main/run_main_module.js:7:1 {
  code: 'ERR_MODULE_NOT_FOUND'
}
```

guybedford: Does the `/` need to be escaped? `\/` perhaps?

guybedford: Right, the error is what I would have expected, without the `/`.

guybedford: I just tested this out and `@set NODE_OPTIONS=--experimental-modules --no-warnings --loader "//%JSPM_PATH%node_modules\\jspm\\node_modules\\@jspm\\resolve\\loader.mjs"` seems to work for me here.

Perhaps try that out and let me know if that fixes this issue for you?

dy: @guybedford that seems to resolve path fine now, but still getting that nodejs error
```
$ jspm bin http-server
internal/modules/esm/default_resolve.js:59
  let url = moduleWrapResolve(specifier, parentURL);
            ^

Error: Cannot find module '' imported from C:\projects\spect\
    at Loader.resolve [as _resolve] (internal/modules/esm/default_resolve.js:59:13)
    at Loader.resolve (internal/modules/esm/loader.js:70:33)
    at Loader.getModuleJob (internal/modules/esm/loader.js:143:40)
    at Loader.import (internal/modules/esm/loader.js:127:28)
    at internal/process/esm_loader.js:44:43
    at Object.exports.initializeLoader (internal/process/esm_loader.js:51:5)
    at initializeESMLoader (internal/bootstrap/pre_execution.js:373:9)
    at prepareMainThreadExecution (internal/bootstrap/pre_execution.js:52:3)
    at internal/main/run_main_module.js:7:1 {
  code: 'ERR_MODULE_NOT_FOUND'
}
```
as if node just ignores the loader

guybedford: @dy just getting back to this now, apologies for the delay. If you are still around to debug it I would value being able to get to the root of this.

I would suggest editing the file at `C:\Program Files\nodejs\node_modules\jspm\node_modules\@jspm\resolve\loader.mjs` directly to include some logging to see what is going on here.

Try adding a `console.log('init')` at the top of that file, and then in the main `resolve` function a log noting something like `console.log('Resolving ' + name + ' to ' + parentUrl)`. That should help narrow down this latest issue.

I've gone ahead and pushed the escaping fix in already now in https://github.com/jspm/jspm-cli/commit/34c329d4054123a4076b0985566861f74c67edde (release to follow in due course).

dy: I'm not around anymore unfortunately, but I tried doing that. The `console.log('init')` in `C:\Program Files\nodejs\node_modules\jspm\node_modules\@jspm\resolve\loader.mjs` never triggered, as if file was just silently ignored by node.

guybedford: Alternatively, you can test this manually with the command:

```
set NODE_OPTIONS=--experimental-modules --loader C:\Program Files\nodejs\node_modules\jspm\node_modules\@jspm\resolve\loader.mjs
node path\to\jspm_packages\http-server@0.11.1\bin\http-server
```

Note the `path\to\` part would need to be the path to the jspm_packages folder.

guybedford: And thanks for testing this out again - it's just that I don't have any ability to replicate the issue here due to it not being standard for the global node_modules to be in Program Files, so really not sure how to help apart from simply giving you the debugging steps directly effectively.

Issue #2490, opened 2019-07-24
jspm link `-i` flag


guybedford:
An `--ignore-uninstalled` or `-i` flag for link could be useful for cases where the provided dependency name to link into does not already exist as an installed package.

For example - 

```
jspm install x y
jspm link ../localx ../localy ../localz
```

would link the `localx` folder against `jspm_packages/[registry]/x@[version]` and for `y`, but because we didn't already install `z` it will now install it as a new dependency in the package.json file.

With the `-i` flag:

```
jspm link -i ../localx ../localy ../localz
> Note: Skipping link of z as it is not already installed
```

we'd automatically skip linking `z` as it is not already installed.

Issue #2489, opened 2019-07-24
jspm link workflow polishing, resolution to be lockfile-specific


guybedford:
This is a tracking issue for the feature that:

```
jspm link ../path
```

should not reflect in the package.json (since the nature of linking is going against the package.json intent for instead following a local intent). Rather it should only represent the link in the lockfile.

In addition `jspm unlink` should properly revert to the previous package.json intent.

fusionstrings:
I think jspm link should not be necessary if somehow monorepo can be detected the way `workspaces` key in package.json is used in yarn.
Please please please

guybedford:
@fusionstrings it could certainly be possible to support the `workspaces` package.json field just like Yarn, but I would want to do it in a way that would effectively be built out of these primitives anyway... because these also open up other opportunities for monorepos, like a monorepo model where the base-package is still a publishable package, with GitHub submodules for the linkage (my ideal monorepo pattern for jspm itself).

So my focus for now continues to be getting the `jspm link` primitive right, then we can easily add the `"workspaces"` sugar in future for sure.

fusionstrings:
Certainly I trust your vision more than anything. I do remember numerous discussions with you on what exactly is JSPM 2 when essentially it strips down the features from jspm0.16 and 0.17. 

And you have been right all along.

I would eagerly wait for this feature to be shipped.

Issue #2461, opened 2019-05-14
what's the best practice for referencing resource files


pigcan:
It is a fresh idea that it is can help us to build a future without webpack just as `@pika/web` does. 

Flowing the guide I understood  your ideas and proposals. As a font-end solution, maybe the guide missed something.Yes they are resource files styles, images etc. As webpack treats all types of files as a module, we can easily import a non-JS file into our JS files. There seems to be no way out under this structure unless we introduce `loader` again.

I appreciate your hard work and this project inspires me a lot!

guybedford:
Glad to hear you're finding the approaches interesting.

At the time Webpack and jspm 1 were created, we thought we were going to have loader hooks in the browser that would support the use cases of importing assets.

Instead, it looks like we will be getting support for specific web assets as their own web specs. Eg importing CSS files is a current spec. Importing JSON files is its own spec. And similarly for importing HTML and WASM.

Currently both JSON and WASM imports are moving very fast in their support and may even be supported in Node and browsers by the end of the year.

So we would definitely support these cases when that is clear.

For assets that don't have any spec proposal, the workflow that would be advised is instead a file load.

So eg:

```js
import fetch from 'universal-fetch';
fetch(new URL('./file.anytype', import.meta.url));
```

The build tooling would then detect the use of `import.meta.url` and see that it is referencing a real file, and copy that file to the build folder.

The action items there are thus - JSON imports (tracking issue - https://github.com/jspm/jspm-cli/issues/2443), WASM imports (just landing into Node.js master right now! https://github.com/jspm/jspm-cli/issues/2465), and asset emission from `import.meta.url` analysis (https://github.com/jspm/rollup-plugin-jspm/issues/6).

CSS imports and HTML imports are a little ways off though still due to their specs being very early.

Hope that gives some ideas, and do let me know what you are interested in further here.

pigcan:
Thanks so much for your reply.

In my opinion 

```js
import fetch from 'universal-fetch';
fetch(new URL('./file.anytype', import.meta.url));
```

it is a little bit unfriendly to UX/DX. For the loading of other files, we all need to implement it ourselves. As you said CSS imports and HTML imports are a little ways off though still due to their specs being very early. 

At the same time, I feel that in the real cases we usually need to process this file (e.g. transform) instead of simply copying it.

So my proposal is  back the concept of `loader` but in different ways.

----- It is my raw idea. -----

I think we can make a step forward based on `fetch` to introduce a more high level API.

[...]

PR #2491, opened 2019-07-24
Support multi-project commands

guybedford:
This attempts a new sort of spin on the Yarn workspaces concept.

Currently `jspm install x -p ./project/path` can be used to run a jspm command from a project that is not the current `cwd`.

So the idea is quite simple - _allow the `-p` flag to accept multiple projects, and that way we can operate on many projects at once, effectively getting the monorepo tools out of a simple primitive_.

For example, consider having `packages/[pkgname]` containing a bunch of packages in a monorepo pattern. Then the following workflows would apply:

1. `jspm install -p packages/*` would install dependencies in all packages.
2. `jspm publish -p packages/*` would publish all packages. Some kind of publish flag to only publish new versions could be used to get a proper monorepo workflow out of this.
3. `jspm run test -p packages/*` would run the package.json "scripts" for `"test"` in each of the packages.

It could be good to work towards a documentation section covering these techniques comprehensively as we polish the various workflows.

The current limitation with linking the packages together is the following - 

In order to link the packages between eachother, we can use:

```
jspm link packages/* -p packages/*`
```

What this does, is updates the `jspm_packages` symlinks in each of the packages to reflect the local versions.

This works fine, but unfortunately has two issues:

1. It would change the package.json to include `"dependencies": { "dep": "link:../dep" }`, which is not what we would want - rather we would just want the linked package to be represented in the package lock file. I've created a tracking issue for this in https://github.com/jspm/jspm-cli/issues/2489.
2. It would install as dependencies all packages in eachother, whereas what we really want is a link flag to say _link only if it is already a dependency_. I've created a tracking issue for this in https://github.com/jspm/jspm-cli/issues/2490.

Rather than trying to do everything at once in this PR, will tackle the above in separate follow-ups.

This is very much an initial PR to demonstrate the concepts - would be interested to hear what people think further about it too.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -48,21 +48,23 @@ function readTargetEquals (installArg: string) {
   return { name, target };
 }
 
-export default async function cliHandler (projectPath: string, cmd: string, args: string | string[]) {
+export default async function cliHandler (projectPaths: string[], cmd: string, args: string[]) {
+  if (typeof projectPaths === 'string')
+    projectPaths = [projectPaths];
   if (typeof args === 'string')
-    args = args.split(' ');
+    args = (<string>args).split(' ');
   
   let setProjectPath = false;
-  let project: api.Project;
+  const projects: api.Project[] = [];
   try {
     let userInput = true, offline = false, preferOffline = false;
     // first read global options
-    for (let i = 0; i < args.length; i++) {
+    outer: for (let i = 0; i < args.length; i++) {
       const arg = args[i];
       switch (arg) {
         case '-y':
         case '--skip-prompts':
-          (<string[]>args).splice(i--, 1);
+          args.splice(i--, 1);
           ui.setUseDefaults(true);
           userInput = false;
         break;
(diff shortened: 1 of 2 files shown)


fusionstrings:
Thanks a lot for getting started with this. This is a very critical feature I am looking forward to.
I have been using yarn workspaces and I like it a lot. Similarity in implementation would be preferred.

- Running `jspm <command> -p ./project/path` is nice and now I am confused if I would like `jspm install x -p <package-name>` more.

- I would definitely wish to not needing to do manual linking and unlinking. I would prefer the yarn way. 
    - Install from local if package@version found locally else download from remote.
    - Publishing not required
    - Packages are resolved by package name.
- Being able to resolve to all the files from package even non js or non main ones. 

Apart from expecting the solution to be more aligned with yarn implementation, I am also curious to see how monorepo solution shapes up for following use cases.

- non `package.json` environments, i.e. Deno.
- non `node_modules` environments i.e. yarn pnp. In this case jspm will not be able to fallback on node_modules folder.

guybedford:
Thanks for the feedback here @fusionstrings 

Points definitely taken on the Yarn-level usability for the final workflow here, as mentioned this is just a first step and further work will be needed.

> Being able to resolve to all the files from package even non js or non main ones.

I'm not sure I follow this point, is this about using the jspm resolver API?

Deno support may well be possible for jspm in future. My hope is to support all JS runtimes, provided they have the necessary resolver hook. Yarn Pnp is actually a direct competitor to jspm in terms of changing the resolver. I believe the jspm approach to be superior as you can still see your packages in jspm_packages, but they are symlinked to the global cache so you don't have the disk size bloat. So you get the benefits without the costs of the hacks (eg hacking fs, moving sources to a hidden global cache folder instead of keeping them project-level inspectable / debuggable). Auto install may well be a future feature, but for now I don't see that as a major workflow problem when developing libraries or apps.

fusionstrings:
Thanks for all the hard work Sir. You inspire and educate me a lot.
With following 

> Being able to resolve to all the files from package even non js or non main ones.

I meant use cases where one wants to access non javascript files from a package i.e. css, icons etc.

Even though jspm doesn't intend to support non standard js behaviours but valid use cases are where other files are accessed in standard way but from one of the packages. 

Like https://github.com/jspm/jspm-cli/issues/2461 but the path could be 
```js
fetch(new URL('@scope/package/file.anytype', import.meta.url));
```

May be when using some plugin this can be supported as well

```js
import @scope/package/file.anytype
```

So I suppose I am referring to jspm resolver API but don't know if this has already been taken care of.

PR #2492, opened 2019-07-26
Add mkdirp as explicit dependency in package.json

vcuhar:
Transitive dependencies do not work well with package managers that do not flatten modules (i.e. pnpm)

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "https-proxy-agent": "^2.1.0",
     "liftoff": "^2.3.0",
     "micromatch": "^3.1.10",
+    "mkdirp": "^0.5.1",
     "nan": "^2.13.2",
     "ncp": "^2.0.0",
     "netrc": "^0.1.4",


guybedford:
Can you further explain the issue this is aiming to fix? Is it that the wrong version of `mkdirp` is resolving, or that it is not resolving at all?

vcuhar:
When using package manager that does flatten node_modules, one may accidentally require packages that are not declared as dependencies in package.json.
See [pnpm FAQ](https://pnpm.js.org/docs/en/faq.html#pnpm-does-not-work-with-your-project-here) for details.

guybedford:
Ah, I see it's an undeclared dependency - yes that's a bug, thank you for posting!

vcuhar:
btw, there are eslint / TSLint rules for that.
i.e. https://palantir.github.io/tslint/rules/no-implicit-dependencies/

PR #2493, opened 2019-07-31
Handle scope normalization as base-relative

guybedford:
This updates the scope handling in jspm to match the latest SystemJS 5.0.0 and es-module-shims in treating scopes as base-relative instead of scope-relative in the import maps.


guybedford:
Released in 2.0.0-beta.7.

PR #2511, opened 2020-03-24
Fix Object.assign({}, string)

GongT:
cause env.https_proxy not work

### DIFF
--- a/src/install/fetch.ts
+++ b/src/install/fetch.ts
@@ -269,7 +269,7 @@ export default class FetchClass {
         if (credentials.proxy) {
           if (!existingAgents)
             proxyAgents.set(credentials.proxy, existingAgents = []);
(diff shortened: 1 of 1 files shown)


src/install/fetch.ts
```diff
@@ -269,7 +269,7 @@ export default class FetchClass {
         if (credentials.proxy) {
           if (!existingAgents)
             proxyAgents.set(credentials.proxy, existingAgents = []);
-          existingAgents.push(agent = new HttpsProxyAgent(Object.assign({}, credentials.proxy)));
+          existingAgents.push(agent = new HttpsProxyAgent(Object.assign({}, agentOptions)));
```
guybedford: Amazing thank you for spotting this.

guybedford (approved):
Thanks for the contribution!

Just to warn you the jspm 2 release is currently on the verge of being abandoned due to a lack of user interest.

PR #2512, opened 2020-03-25
Fix install scoped package with binaries

jbanety:
When I install a scoped package with binaries I'll get this error :
```bash
$ jspm install  @mapbox/geojson-rewind
err (jspm) Internal Error: Unhandled promise rejection.
Error: ENOENT: no such file or directory, open '[...]/jspm_packages/.bin/@mapbox/geojson-rewind'
```
jspm install command don't stripe scope in path so the destination folder doesn't exists.
My PR is to remove the scope with a regexp before writing file.

### DIFF
--- a/src/install/index.ts
+++ b/src/install/index.ts
@@ -709,7 +709,7 @@ export class Installer {
         this.binFolderChecked = true;
       }
       await Promise.all(Object.keys(config.bin).map(p => 
-        writeBinScripts(binDir, p, resolvedPkgName.replace(':', path.sep) + path.sep + config.bin[p])
+        writeBinScripts(binDir, p.replace(/(@.*\/)/, ""), resolvedPkgName.replace(':', path.sep) + path.sep + config.bin[p])
       ));
     }
   }


guybedford:
Thanks for the PR. This is originally generated in https://github.com/jspm/jspm-cli/blob/master/src/install/package.ts#L449 so I think we should add the scope check there, which is simply then `name[0] === '@' ? name.slice(name.indexOf('/')) : name` or something along those lines.

Just to warn you, due to a lack of user interest in jspm 2, the entire version is considered deprecated and will be abandoned soon.

jbanety:
Hi @guybedford,
I'm sad to hear that. 
It was very promising for the future of web app.
What will be the alternatives?

guybedford:
I'm working on a jspm 3 version which will entirely move away from trying to recreate local package management, since it seems that combination of approach is too complex for most users to grasp.

jbanety:
How Can I help you?

Le ven. 27 mars 2020 à 17:38, Guy Bedford <notifications@github.com> a
écrit :

guybedford:
If you're interested in trying out / contributing to jspm 3, it should be ready for further testing within the next month.

jbanety:
👍
I have an internal project on which I will work during covid-19 confinment.

Le ven. 27 mars 2020 à 18:42, Guy Bedford <notifications@github.com> a
écrit :

fusionstrings:
What are the ways one can be part of JSPMv3 testing, once and when it's available @guybedford .

guybedford:
@jbanety @fusionstrings I can invite you both to the private testing for jspm 3 when it is ready if you like.

jbanety:
👍 happy to Bell

Cordialement,
Jean-Baptiste.

—

Jean-Baptiste Alleaume
ETD Solutions
06 21 67 84 89

fusionstrings:
> @jbanety @fusionstrings I can invite you both to the private testing for jspm 3 when it is ready if you like.

I will feel extremely privileged @guybedford 👍 .

PR #2514, opened 2020-04-02
docs: Fix simple typo, writeable -> writable

timgates42:
There is a small typo in lib/install/source.js.

Should read `writable` rather than `writeable`.


guybedford:
Hi thanks for the PR. jspm 2 is no longer maintained due to a lack of user interest.

PR #2530, opened 2023-04-10
docs(install): fix usage and description

dasa:
Running `chomp` changed the line endings for `docs.md` to be Unix instead of mixed.

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -24,7 +24,7 @@ By default, JSPM operates on `importmap.json` which is automatically created if 
 ### Usage
   
 ```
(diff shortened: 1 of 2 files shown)


guybedford:
Thank you for spotting this and following-up, this PR got out of date, but the fixes should be in now!

dasa:
The "Usage" still says "link" instead of "install" at: https://github.com/jspm/jspm-cli/blob/main/docs.md#install

guybedford:
@dasa ah right of course, thanks I've pushed up the change, will be updated in the next release.

PR #2531, opened 2023-04-17
feat: supporting changing module preload mode

### LINKED ISSUE #2518: Add support for creating module preloads for only the static dependencies
It currently seems to also create `modulepreload`'s for the dynamic imports. I'd like these to only be loaded as needed.

Bubblyworld:
Fixes #2518. There's an `esm.sh` provider test that's failing currently until
we pull in this downstream fix in the generator:
https://github.com/jspm/generator/pull/295

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -371,9 +371,13 @@ function getCacheMode(flags: Flags): "offline" | boolean {
   return false;
 }
 
-const validPreloadModes = ["static", "dynamic", "no-preloads"];
+const validPreloadModes = ["static", "dynamic"];
 function getPreloadMode(flags: Flags): boolean | string {
-  if (!flags.preload) return "static";
+  if (flags.preload === null || flags.preload === undefined) return false;
+  if (typeof flags.preload === "boolean") {
+    return flags.preload;
+  }
+
   if (!validPreloadModes.includes(flags.preload))
     throw new JspmError(
       `Invalid preload mode "${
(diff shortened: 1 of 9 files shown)


src/utils.ts
```diff
@@ -361,16 +361,37 @@ function getCacheMode(flags: Flags): "offline" | boolean {
...
     );
 
   if (flags.cache === "offline") return "offline";
   if (flags.cache === "online") return true;
   return false;
 }
 
+const validPreloadModes = ["static", "dynamic", "no-preloads"];
+function getPreloadMode(flags: Flags): boolean | string {
+  if (!flags.preload) return "static";
```
guybedford: When no flag is set, we shouldn't do any preloading at all.

src/utils.ts
```diff
@@ -361,16 +361,37 @@ function getCacheMode(flags: Flags): "offline" | boolean {
...
+  if (!validPreloadModes.includes(flags.preload))
```
guybedford, line 381: Can we allow a "boolean" form ie `--preload` without a value? Perhaps we need to preparse this in the CLI handling?

src/utils.ts
Bubblyworld: Mmm, think I misunderstood what you meant by defaulting to static. You mean default to static only if the flag is present at all?

guybedford: Yes exactly, preloading shouldn't be blanket enabled by default.

PR #2539, opened 2023-06-17
docs: show available providers in options

canadaduane:
Show a complete list of available providers that can be used as parameter to the --provider install option.


guybedford:
Nice, thanks for the PR!

PR #2542, opened 2023-08-14
(chore): Bump @jspm/generator@1.1.10

JayaKrishnaNamburu:


guybedford:
Not sure what is with the test failure here, I think we will need to look into that.

JayaKrishnaNamburu:
Tests are fixed now, they are just related to `es-module-shims` and `es-module-lexer` version bumps from the `@jspm/generator` 👍

guybedford (approved):
Awesome thanks for fixing that

PR #2544, opened 2023-09-22
(feat): Add build command by wrapping rollup

JayaKrishnaNamburu:
This PR aims to introduce the `build` command. Which will work in 2 modes. Basically the command wraps the `rollup` Node API with some defaults to run in simple mode. Where, we define some standard defaults and works with some projects out of the box.

```sh
jspm build --entry app.js --output build/
```

Then there is an another mode, where users can extend the behaviour with custom rollup setups. And the wrapper loads the config from the URL and starts the build.

```sh
jspm build --build-config build.mjs
```

is a simple rollup file. Which can have any defaults in it. Example

```js
// build.mjs
import { defineConfig } from "rollup";
import banner from "rollup-plugin-banner";
import { babel } from "@rollup/plugin-babel";
import { nodeResolve } from "@rollup/plugin-node-resolve";

export default defineConfig({
  input: "./app.js",
  plugins: [
    nodeResolve(),
    babel({
      presets: ["@babel/preset-react"],
    }),
    banner.default("Running rolup with a wrapper around with jspm cli"),
  ],
  output: {
    file: "./build.js",
  },
});

```

Which generates a output of

```js
// Running rolup with a wrapper around with jspm cli
import react from 'react';

console.log('This is from utils file');

console.log(react);
const App = () => {
  return /*#__PURE__*/React.createElement("div", null, "Hello World");
};

export { App };

```

The `importmap.json` processing is not handled yet. But once done, it can bundle `react` from `jspm`  CDN and include in the build.

### DIFF
--- a/chompfile.toml
+++ b/chompfile.toml
@@ -10,7 +10,7 @@ deps = ['dist/cli.js', 'docs.md']
 [[task]]
 target = 'dist/cli.js'
 deps = ['src/**/*.ts', 'npm:install']
-run = 'esbuild src/cli.ts --bundle --platform=node --external:@jspm/generator --external:ora --external:picocolors --external:@babel/core --format=esm --outfile=$TARGET'
+run = 'esbuild src/cli.ts --bundle --platform=node --external:fsevents --external:@jspm/generator --external:ora --external:picocolors --external:@babel/core --format=esm --outfile=$TARGET'
 
 [[task]]
 name = 'docs'
--- a/src/build/index.ts
+++ b/src/build/index.ts
@@ -1,0 +1,60 @@
+import path from "node:path";
+import process from "node:process";
+import fs from "node:fs/promises";
+import { type RollupOptions, rollup } from "rollup";
+
+import { JspmError, exists } from "../utils";
+import type { Flags } from "../types";
+import { RollupImportmapPlugin } from "./rollup-importmap-plugin";
+
+export default async function build(flags: Flags) {
+  if (!flags.entry && !flags.buildConfig) {
+    throw new JspmError(`Please provide entry for the build`);
+  }
+
+  let buildConfig: RollupOptions;
+  let outputOptions: RollupOptions["output"];
+
+  if (flags.entry) {
+    const entryPath = path.join(process.cwd(), flags.entry);
+    if ((await exists(entryPath)) === false) {
+      throw new JspmError(`Entry file does not exist: ${entryPath}`);
+    }
+    buildConfig = {
+      input: entryPath,
+      plugins: [RollupImportmapPlugin(flags)],
+    };
+  }
+
+  if (flags.buildConfig) {
+    const buildConfigPath = path.join(process.cwd(), flags.buildConfig);
+    if ((await exists(buildConfigPath)) === false) {
+      throw new JspmError(
+        `Build config file does not exist: ${buildConfigPath}`
+      );
+    }
+    const rollupConfig = await import(buildConfigPath)
+      .then((mod) => mod.default)
+      .catch((err) => {
+        throw new JspmError(`Failed to load build config: ${err}`);
+      });
+
+    if ("output" in rollupConfig) {
+      outputOptions = rollupConfig.output;
+    }
+
+    buildConfig = {
+      ...rollupConfig,
+      plugins: [...(rollupConfig?.plugins || []), RollupImportmapPlugin(flags)],
+    };
+  }
+
+  const builder = await rollup(buildConfig);
+  const result = await builder.generate({ format: "esm", ...outputOptions });
+
+  for (const file of result.output) {
+    const outputPath = path.join(process.cwd(), file.fileName);
+    const content = file.type === "asset" ? file.source : file.code;
+    await fs.writeFile(outputPath, content, "utf-8");
+  }
+}
--- a/src/build/index.ts
+++ b/src/build/index.ts
@@ -7,27 +7,35 @@ import { JspmError, exists } from "../utils";
 import type { Flags } from "../types";
 import { RollupImportmapPlugin } from "./rollup-importmap-plugin";
 
-export default async function build(flags: Flags) {
-  if (!flags.entry && !flags.buildConfig) {
+export default async function build(entry: string, options: Flags) {
+  if (!entry && !options.buildConfig) {
     throw new JspmError(`Please provide entry for the build`);
   }
 
   let buildConfig: RollupOptions;
   let outputOptions: RollupOptions["output"];
 
-  if (flags.entry) {
-    const entryPath = path.join(process.cwd(), flags.entry);
+  if (entry) {
+    if (!options.buildOutput) {
+      throw new JspmError(`Build output is required when entry is provided`);
+    }
+
+    const entryPath = path.join(process.cwd(), entry);
     if ((await exists(entryPath)) === false) {
       throw new JspmError(`Entry file does not exist: ${entryPath}`);
     }
     buildConfig = {
       input: entryPath,
-      plugins: [RollupImportmapPlugin(flags)],
+      plugins: [RollupImportmapPlugin(options)],
+    };
+
+    outputOptions = {
+      file: path.join(process.cwd(), options.buildOutput),
     };
   }
 
-  if (flags.buildConfig) {
-    const buildConfigPath = path.join(process.cwd(), flags.buildConfig);
+  if (options.buildConfig) {
+    const buildConfigPath = path.join(process.cwd(), options.buildConfig);
     if ((await exists(buildConfigPath)) === false) {
       throw new JspmError(
         `Build config file does not exist: ${buildConfigPath}`
@@ -45,7 +53,10 @@ export default async function build(flags: Flags) {
 
     buildConfig = {
       ...rollupConfig,
-      plugins: [...(rollupConfig?.plugins || []), RollupImportmapPlugin(flags)],
+      plugins: [
+        ...(rollupConfig?.plugins || []),
+        RollupImportmapPlugin(options),
+      ],
     };
   }
 
--- a/src/build/index.ts
+++ b/src/build/index.ts
@@ -1,6 +1,5 @@
 import path from "node:path";
 import process from "node:process";
-import fs from "node:fs/promises";
 import { type RollupOptions, rollup } from "rollup";
 
 import { JspmError, exists } from "../utils";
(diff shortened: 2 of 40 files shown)


JayaKrishnaNamburu:
Made some resolution changes with the `rollup-plugin` now. It can download the deps and bundle them too. 
The command

```sh
jspm build --build-config build-config.mjs
```

downloads all the dependencies that are defined in the `importmap.json` and bundles them along.

JayaKrishnaNamburu:
Crazy how this wraps the whole [vite-plugin-jspm](https://github.com/jspm/vite-plugin-jspm) functionality 😄

chompfile.toml
```diff
@@ -10,7 +10,7 @@ deps = ['dist/cli.js', 'docs.md']
 [[task]]
 target = 'dist/cli.js'
 deps = ['src/**/*.ts', 'npm:install']
-run = 'esbuild src/cli.ts --bundle --platform=node --external:@jspm/generator --external:ora --external:picocolors --external:@babel/core --format=esm --outfile=$TARGET'
+run = 'esbuild src/cli.ts --bundle --platform=node --external:fsevents --external:@jspm/generator --external:ora --external:picocolors --external:@babel/core --format=esm --outfile=$TARGET'
```
guybedford, line 13: We should move to a self-build at some point if we can...!

src/build/rollup-importmap-plugin.ts
```diff
@@ -0,0 +1,51 @@
...
+  const generator = await getGenerator({ ...flags, freeze: true });
+  await generator.install();
+
+  return {
+    name: "rollup-importmap-plugin",
+    resolveId: async (id: string, importer: string) => {
+      if (importer?.startsWith("http") && id?.startsWith(".")) {
+        const proxyPath = new URL(id, importer).toString();
+        return { id: proxyPath, external: false };
+      }
```
guybedford: Can you share the reason for why this condition is needed?

src/build/rollup-importmap-plugin.ts
```diff
@@ -0,0 +1,51 @@
...
+        const url = new URL(id);
+
+        if (url.protocol === "file:") {
+          /*
+            This is a hack and need to be replaced with a proper solution.
+          */
+          const filePath =
+            path.extname(url.pathname) === ""
+              ? `${url.pathname}.js`
+              : url.pathname;
```
guybedford, line 52: Make sure to use `pathToFileURL` from Node.js `import { pathToFileURL } from 'node:url'` here.

src/build/rollup-importmap-plugin.ts
```diff
@@ -0,0 +1,51 @@
...
+      }
+    },
+    load: async (id: string) => {
+      try {
+        const url = new URL(id);
+
+        if (url.protocol === "file:") {
+          /*
+            This is a hack and need to be replaced with a proper solution.
+          */
```
guybedford: Can you share why the extension adding is necessary?

src/build/rollup-importmap-plugin.ts
```diff
@@ -0,0 +1,51 @@
...
+    name: "rollup-importmap-plugin",
+    resolveId: async (id: string, importer: string) => {
+      if (importer?.startsWith("http") && id?.startsWith(".")) {
+        const proxyPath = new URL(id, importer).toString();
+        return { id: proxyPath, external: false };
+      }
+
+      try {
+        const resolved = generator.importMap.resolve(id, importer);
+        return { id: resolved };
```
guybedford, line 37: We should probably mark Node.js / Deno builtins etc as external over here.

src/build/rollup-importmap-plugin.ts
```diff
@@ -0,0 +1,51 @@
...
+
+          return await fs.readFile(filePath, "utf-8");
+        }
+
+        if (url.protocol === "https:") {
+          const response = await fetch(id);
+          return await response.text();
+        }
+      } catch (err) {
+        throw new JspmError(`\n Failed to resolve ${id}: ${err.message} \n`);
```
guybedford: This should probably specifically be an error that reads "unsupported protocol".

src/cli.ts
```diff
@@ -88,6 +91,16 @@ const freezeOpt: opt = [
   { default: false },
 ];
 const silentOpt: opt = ["--silent", "Silence all output", { default: false }];
+const buildOpt: opt = [
+  "--entry <file>",
```
guybedford: Instead of making this a flag, we should just make these the arguments to the build. Also because Rollup supports multiple entry points, we should support multiple arguments too.

src/build/index.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export default async function build(flags: Flags) {
+  if (!flags.entry && !flags.buildConfig) {
+    throw new JspmError(`Please provide entry for the build`);
+  }
+
+  let buildConfig: RollupOptions;
+  let outputOptions: RollupOptions["output"];
+
+  if (flags.entry) {
```
guybedford: What happens if entry is not provided? Surely we always need one?

guybedford (commented):
Excellent work, very exciting to see this one. Can you add some tests? Let me know if you need help getting them going.

guybedford:
Also documentation should be added in https://github.com/jspm/jspm-cli/tree/main/docs, which would then get added to the site.

JayaKrishnaNamburu:
Thanks for the review @guybedford, good to see that the PR was in the right direction 😄. Will make the necessary changes, and see how we can integrate this with other commands. Like we have `traceInstall` available with `jspm installl` that updates `.html` files. 

But if users are bundling the deps too. We don't need to add `importmap` anymore. So, just trying to list use-cases like those. And see how we can streamline the workflows and connect these commands more together 👍

src/build/rollup-importmap-plugin.ts
JayaKrishnaNamburu: When users use imports without extensions like

```js
import { util  } from './utils'
```

The `load` fails to resolve this `./utils` I am not sure, if rollup allows to return these and let the other plugins like `rollup-plugin-node-resolve` handle the import.

src/build/rollup-importmap-plugin.ts
JayaKrishnaNamburu: Not needed, it was to solve relative imports that comes from `https` content. But, this is solved by passing `parentURL` to the generator. Removed these lines now 👍

chompfile.toml
JayaKrishnaNamburu: Regarding this, i have some doubts. On what's the best possible way

- Show we just mark `rollup` as a `peerDependency` and `options`. Since users who don't use `jspm build` don't need to install `rollup` ?
- If we define `rollup` as direct dependency instead of `peer` or `optional`. What happens if users install their own version of `rollup` while defining the `config.mjs` file. Which version of rollup get's picked. The one from the cli or the one coming from the users project when they run `jspm build` ?

src/build/rollup-importmap-plugin.ts
JayaKrishnaNamburu: Now i am wondering should we even handle the imports without extensions 🤔

src/build/index.ts
JayaKrishnaNamburu: We don't need the `--entry` all the time. As they can export `input` from the `config` file. So, when build config is passed. Rollup takes the entry from it.

chompfile.toml
JayaKrishnaNamburu: Here is one from tests, where users don't need to install rollup. And just export the config alone. 
https://github.com/jspm/jspm-cli/blob/fa9c2d676710ca2b3c55eac5ecca9afc68539e1e/test/fixtures/scenario_build_app/rollup-config.mjs

src/build/rollup-importmap-plugin.ts
guybedford: In Node.js this only works when the package is loaded as CommonJS, not ES modules. And JSPM generator follows these rules, so I think we should do that.

src/build/index.ts
guybedford: Thanks for clarifying. I would still prefer to make these optional arguments over a flag.

JayaKrishnaNamburu:
@guybedford  made the final changes, and added the documentation for the same. Please check if need any changes/we can merge and release it.

src/cli.ts
```diff
@@ -88,6 +91,16 @@ const freezeOpt: opt = [
   { default: false },
 ];
 const silentOpt: opt = ["--silent", "Silence all output", { default: false }];
+const buildConfigOpt: opt = [
+  "--build-config <file>",
```
guybedford: Let's make this just `--config`.

src/cli.ts
```diff
@@ -88,6 +91,16 @@ const freezeOpt: opt = [
...
+  "Path to a rollup config file",
+  {},
+];
+const buildOutputOpt: opt = [
+  "--build-output <file>",
```
guybedford: Rollup outputs a directory not a single file. I think it's important to make the directory case the default. Also `--output / -o <dir>` . Let's remove the file case entirely.

src/build/index.ts
```diff
@@ -0,0 +1,71 @@
...
+    if ((await exists(entryPath)) === false) {
+      throw new JspmError(`Entry file does not exist: ${entryPath}`);
+    }
+    buildConfig = {
+      input: entryPath,
+      plugins: [RollupImportmapPlugin(options)],
+    };
+
+    outputOptions = {
+      file: path.join(process.cwd(), options.buildOutput),
```
guybedford: Please let's use the `dir` option to support multiple entry points and dynamic import by default.

guybedford (commented):
Thanks for working on the feedback here. I've left another couple of change requests. With those I think we could then approve and land and release. Ping me on discord if you want to discuss further at all.

JayaKrishnaNamburu:
Done with the directory changes 👍

src/build/rollup-importmap-plugin.ts
```diff
@@ -0,0 +1,65 @@
...
+        if (url.protocol === "deno:" || url.protocol === "node:") {
+          return { id, external: true };
+        }
+      }
+
+      try {
+        const resolved = generator.importMap.resolve(id, importer);
+        return { id: resolved };
+      } catch (err) {
+        return { id, external: true };
```
guybedford, line 42: We probably do want to provide module not found errors. And then have some handling for externals. At the very least perhaps we should use warnings?

I'm happy for us to follow-up on this though and merge now if you prefer.

guybedford (approved):
Thanks, looks good down to the externals handling versus resolution errors.

src/build/rollup-importmap-plugin.ts
JayaKrishnaNamburu: Should we through a console error for now and then allow users to provide a option to install the deps automatically down the line ?

Adding a `console.warn` for now 👍

JayaKrishnaNamburu:
Should i release `3.1.0` with this change ?

guybedford:
Yes please!

Issue #2547, opened 2023-10-30
Fix Windows Tests


guybedford:
The tests for the CLI are currently failing on Windows.

guybedford:
Resolved in https://github.com/jspm/jspm-cli/pull/2548.

PR #2548, opened 2023-10-30
ci: add Windows CI tests

guybedford:
Adds a Windows CI job. Currently failing pending https://github.com/jspm/jspm-cli/issues/2547.


guybedford:
For some reason the Windows CI is stalling. I was able to reproduce this locally, but after running it again it worked - so I'm not sure which cache layer might be aiding this / participating in this deadlock. Will need some more investigation.

PR #2555, opened 2023-12-07
fix: Use existing importmap for build by passing `--map` flag

JayaKrishnaNamburu:
Adds `--map` option for `jspm build` command.
While moving the `jspm-vscode` project to use `jspm-cli` workflow. We will generate the `importmap.json` only once. And use it for re-runs or update it only when needed.

So, when the extension is being built. We should use the existing map. Currently the `cli` throws an error because `--map` is not allowed for `build` command.

Usage reference -> https://github.com/jspm/jspm-vscode/blob/refactor-build-workflow/chompfile.toml#L9

Eg

```sh
jspm build --map importmap.json --config rollup-config.mjs
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "3.1.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "@jspm/generator": "^1.1.10",
+        "@jspm/generator": "^1.1.12",
         "cac": "^6.7.14",
         "ora": "^6.3.0",
         "picocolors": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "jspm.js"
   ],
   "dependencies": {
-    "@jspm/generator": "^1.1.10",
+    "@jspm/generator": "^1.1.12",
     "cac": "^6.7.14",
     "ora": "^6.3.0",
     "picocolors": "^1.0.0",
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -293,6 +293,7 @@ Clears the global module fetch cache, for situations where the contents of a dep
 
 cli
   .command("build [entry]", "Build the module using importmap")
+  .option(...mapOpt)
   .option(...buildConfigOpt)
   .option(...buildOutputOpt)
   .action(wrapCommand(build));
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -30,7 +30,11 @@ export const availableProviders = [
   "jsdelivr",
   "skypack",
   "unpkg",
-  "esm.sh",
+  /*
+    Disabling esm.sh provider for now. There is a bug for installing lit.
+    https://github.com/jspm/generator/issues/335
+  */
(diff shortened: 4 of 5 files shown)


guybedford:
Looks good thanks. Any idea about the CI failure? Note that most CLI operations should automatically pick up the local `importmap.json` file without it needing to be provided via an explicit flag. Can we not do the same for the build?

guybedford (approved):

JayaKrishnaNamburu:
Yes, the cli is picking the local build file by default. Needed only if we change the file name to something else than `importmap.json`
https://github.com/jspm/jspm-cli/blob/main/src/utils.ts#L213-L222

JayaKrishnaNamburu:
The CI failures are there for a while, should look into it. Something with assertions

JayaKrishnaNamburu:
Fixed tests for one use-case. There is another use-case that is failing for `jspm install lit` from `esm.sh` now. But i have a hunch it might not be realted to the generator.

Here is the error
```
Error: No './css-tag' exports subpath defined in https://esm.sh/*@lit/reactive-element@1.6.3/ resolving @lit/reactive-element/css-tag imported from https://esm.sh/v135/@lit/reactive-element@1.6.3/X-ZS8q/esnext/reactive-element.mjs.

Error: Scenario "jspm install lit -p esm.sh -e production" failed.
    at runScenario (file:///home/runner/work/jspm-cli/jspm-cli/test/scenarios.ts:1:845)
    at async runScenarios (file:///home/runner/work/jspm-cli/jspm-cli/test/scenarios.ts:1:444) {
  [cause]: AssertionError [ERR_ASSERTION]: undefined == true
      at Object.validationFn (file:///home/runner/work/jspm-cli/jspm-cli/test/providers.test.ts:1:1140)
      at runScenario (file:///home/runner/work/jspm-cli/jspm-cli/test/scenarios.ts:1:790)
      at async runScenarios (file:///home/runner/work/jspm-cli/jspm-cli/test/scenarios.ts:1:444) {
    generatedMessage: true,
    code: 'ERR_ASSERTION',
    actual: undefined,
    expected: true,
    operator: '=='
  }
}
Unable to complete all tasks.
x test/providers.test.ts [24.707770[31](https://github.com/jspm/jspm-cli/actions/runs/7144479899/job/19458208881#step:7:32)s]
x :test:##
x :test
```

The package `@lit/reactive-element` seems to create a export map with `.js` extension. And maybe the us causing the issue when trying to access the same without extensions ?
https://unpkg.com/browse/@lit/reactive-element@2.0.2/package.json

guybedford:
Can we at least disable the test and post a new issue to reenable it? We can't land PRs if CI is failing...

Otherwise looks good to land and thanks for clarifying `importmap.json` is already read by default.

JayaKrishnaNamburu:
Disabled `esm.sh` provider and loged a bug with repro on the generator.
https://github.com/jspm/jspm/issues/2625

PR #2556, opened 2023-12-08
feat: Allow resolutions flag to build command

JayaKrishnaNamburu:
We need to allow resolutions too at times when building. Sharing more details about it here
https://github.com/jspm/jspm-vscode/compare/refactor-build-workflow


guybedford (approved):
Again, we should ensure CI is passing before landing, even if it's a skipped test with an issue posted.

PR #2561, opened 2023-12-22
Update latest `@jspm/generator` and drop the `freeze` option from the cli

JayaKrishnaNamburu:
In the latest version of generators. the `freeze` option in generator is deprecated.
So, we need to reflect the same in the cli. And may be add in future if needed
https://github.com/jspm/generator/releases/tag/2.0.0


guybedford (approved):
Instead of deleting the test, can we mark it as `freeze.test.skip.ts` or similar to get it to be omitted?

JayaKrishnaNamburu:
@guybedford added the freeze file back. And changed the file name. Can you review once again if everything is good please 😄

guybedford:
LGTM

Issue #2575, opened 2024-07-28
HTML linking case importmap input


guybedford:
I just tried linking the following HTML:

```html
<!doctype html>
<script type="importmap">
{
  "imports": {
    "lit": "https://cdn.jsdelivr.net/gh/lit/dist@3.1.4/core/lit-core.min.js"
  }
}
</script>
<script type="module" src="./app.js"></script>
<body>
  <simple-greeting name="World"></simple-greeting>
</body>
```

but when I do: 

```
jspm link app.html -o app.html --integrity
```

it changes the mappings to use `jspm.io` instead.

Instead I need to write:

```
jspm link app.html --map app.html -o app.html --integrity
```

to build integrity against the existing import map.

In the case where there is no `importmap.json` file, we should default to taking the import map from the provided `app.html` link input.

guybedford:
Resolved in #2577.

PR #2577, opened 2024-07-28
feat: link with input HTML map when no importmap.json is present

guybedford:
This updates `jspm link file.html` to use the import map from the HTML file when no `importmap.json` is present, resolving the workflow issue described in https://github.com/jspm/jspm-cli/issues/2575.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -30,7 +30,7 @@ export const cli = cac(c.yellow("jspm"));
 type opt = [string, string, any];
 const mapOpt: opt = [
   "-m, --map <file>",
-  "File containing initial import map",
+  "File containing initial import map (defaults to importmap.json, or the input HTML if linking)",
   {},
 ];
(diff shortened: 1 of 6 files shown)


JayaKrishnaNamburu (approved):
Tested, working as expected now 👍

src/cli.ts
```diff
@@ -31,7 +31,7 @@ type opt = [string, string, any];
 const mapOpt: opt = [
   "-m, --map <file>",
   "File containing initial import map",
-  { default: "importmap.json" },
+  {},
```
JayaKrishnaNamburu, line 34: This needs a documentation change though. `chomp build --force`

```
* `-m, --map` _&lt;file&gt;_                File containing initial import map 
```

But should we still give a hint to users, that the cli picks `importmap.json` if it exists ?

guybedford: Thanks, added.

PR #2648, opened 2025-06-15
Begin work on the downloadAll API

internettrans:
This is a draft, since I would prefer confirmation on the API before continuing. There was discussion in https://github.com/jspm/jspm/issues/2631 and https://github.com/jspm/jspm/issues/2605. I started the implementation with a `checkout()` api but the following comment was unclear to me:

```js
// Output is a map of URLs to local paths
// The import map returned then works against these local paths directly
```

The suggested return value seemed to be two things - a map of URLs to local paths, and an import map. The map of URLs to local paths seemed like it would only be used for download purposes, and exposed unnecessary internals.

@guybedford pls advise which API is most amenable to the jspm generator project.

### DIFF
--- a/generator/src/generator.ts
+++ b/generator/src/generator.ts
@@ -1671,27 +1671,28 @@ export class Generator {
   }
 
   /**
+   * Download all files to disk within the import map to a directory. A rebased import map, with URLs relative
+   * to the output directory is returned.
+   *
+   * Example usage:
+   *
+   * const finalMap = await generator.downloadAll('deps');
    *
    * @param outputDirectory The output directory, resolved relative to process.cwd(), where the import
    * map and its files are stored on the file system. Defaults to 'dist'
-   * @param importMapFileName The name of the import map file stored to disk, within the outputDirectory. Defaults to
-   * main.importmap
-   * @param hostedBaseUrl a fully qualified or relative string path used as base url for all URLs within the import
-   * If not provided, hostedBaseUrl defaults to './'
    */
-  async downloadAll(
-    outputDirectory: string = 'deps',
-    importMapFileName: string = 'main.importmap',
-    hostedBaseUrl: string = './'
-  ) {
+  async downloadAll(outputDirectory: string = 'deps'): Promise<ImportMap> {
     if (!isNode) {
       throw new JspmError('downloadAll is only available in nodejs');
     }
 
-    const fs = await import('node:fs/promises');
-    const path = await import('node:path');
+    const fs = await import(eval('"node:fs/promises"'));
+    const path = await import(eval('"node:path"'));
+
+    const cloudOrigins = new Set<string>();
 
     for (const url in this.traceMap.installer.resolver.traceEntries) {
+      cloudOrigins.add(new URL(url).origin);
       const r = await fetch(url);
       if (!r.ok) {
         throw Error(`Failed to download file from url '${url}'`);
@@ -1702,13 +1703,13 @@ export class Generator {
       await fs.writeFile(filename, text, 'utf-8');
     }
 
-    const map = this.importMap.rebase('https://ga.jspm.io/');
+    // Remove cloud origins from all URLs
+    let finalMap: ImportMap = this.importMap.clone();
+    for (const cloudOrigin of cloudOrigins) {
+      finalMap = finalMap.rebase(cloudOrigin);
+    }
 
-    await fs.writeFile(
-      path.resolve(outputDirectory, importMapFileName),
-      JSON.stringify(map.toJSON(), null, 2),
-      'utf-8'
-    );
+    return finalMap;
   }
 }
 
--- a/generator/src/generator.ts
+++ b/generator/src/generator.ts
@@ -1669,6 +1669,47 @@ export class Generator {
     if (this.combineSubpaths) map.combineSubpaths();
     return map.toJSON();
   }
+
+  /**
+   *
+   * @param outputDirectory The output directory, resolved relative to process.cwd(), where the import
+   * map and its files are stored on the file system. Defaults to 'dist'
+   * @param importMapFileName The name of the import map file stored to disk, within the outputDirectory. Defaults to
+   * main.importmap
+   * @param hostedBaseUrl a fully qualified or relative string path used as base url for all URLs within the import
+   * If not provided, hostedBaseUrl defaults to './'
+   */
+  async downloadAll(
+    outputDirectory: string = 'deps',
+    importMapFileName: string = 'main.importmap',
+    hostedBaseUrl: string = './'
+  ) {
+    if (!isNode) {
+      throw new JspmError('downloadAll is only available in nodejs');
+    }
+
+    const fs = await import('node:fs/promises');
+    const path = await import('node:path');
+
+    for (const url in this.traceMap.installer.resolver.traceEntries) {
+      const r = await fetch(url);
+      if (!r.ok) {
+        throw Error(`Failed to download file from url '${url}'`);
+      }
+      const text = await r.text();
+      const filename = path.resolve(outputDirectory, '.' + new URL(url).pathname);
+      await fs.mkdir(filename.slice(0, filename.lastIndexOf('/')), { recursive: true });
+      await fs.writeFile(filename, text, 'utf-8');
+    }
+
+    const map = this.importMap.rebase('https://ga.jspm.io/');
+
+    await fs.writeFile(
+      path.resolve(outputDirectory, importMapFileName),
+      JSON.stringify(map.toJSON(), null, 2),
+      'utf-8'
+    );
+  }
 }
 
 export interface LookupOptions {
--- a/generator/.gitignore
+++ b/generator/.gitignore
@@ -3,3 +3,4 @@ node_modules
 lib
 yarn.lock
 .vscode
+fixtures
--- a/generator/test/api/download.test.js
+++ b/generator/test/api/download.test.js
@@ -1,0 +1,8 @@
+import { Generator } from '@jspm/generator';
+
+const generator = new Generator();
+await generator.install('react');
+await generator.install('dayjs');
+await generator.install('rxjs');
+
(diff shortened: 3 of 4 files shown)


generator/src/generator.ts
```diff
@@ -1669,6 +1669,47 @@ export class Generator {
...
+    hostedBaseUrl: string = './'
+  ) {
+    if (!isNode) {
+      throw new JspmError('downloadAll is only available in nodejs');
+    }
+
+    const fs = await import('node:fs/promises');
+    const path = await import('node:path');
+
+    for (const url in this.traceMap.installer.resolver.traceEntries) {
```
internettrans, line 1694: I'm not sure if this captures all of the files that could possibly be used. I'm not sure if it's possible for jspm-generator import maps to load files that then load additional modules that are not contained within the import map.

I have [another implementation of this](https://github.com/single-spa/import-map-microfrontend-deps/blob/62a9f6cc6c6a93e8752c94d5c6b5cbd6bfd9ee09/src/import-map-microfrontend-deps.ts#L101) within the single-spa ecosystem, and used the `integrity` field of the import map to obtain the urls to download, rather than the `resolver` within jspm generator.

This is my first contribution to the project - looking forward to feedback on it.

guybedford:
Hey @jolyndenning thanks for taking a stab at this one. I'd be happy to walk you through some design concepts here further. Perhaps we can arrange a discussion sometime to talk through how to achieve this? There's certainly many ways to skin this one, and I've gone back and fourth myself on design directions a lot here, but right now I'm thinking of moving the JSPM CLI back in the direction of supporting a `jspm_packages/registry/layer/package@version/` direction per the original design but with `layer` added to account for build layers which was the original design flaw.

The checkout workflow is a nice one to solve in the process, but yeah perhaps it is more of a secondary feature than the primary one.

It would be great if you want to join the JSPM Discord to discuss further here.

generator/src/generator.ts
```diff
@@ -1669,6 +1669,47 @@ export class Generator {
...
+   * @param outputDirectory The output directory, resolved relative to process.cwd(), where the import
+   * map and its files are stored on the file system. Defaults to 'dist'
+   * @param importMapFileName The name of the import map file stored to disk, within the outputDirectory. Defaults to
+   * main.importmap
+   * @param hostedBaseUrl a fully qualified or relative string path used as base url for all URLs within the import
+   * If not provided, hostedBaseUrl defaults to './'
+   */
+  async downloadAll(
+    outputDirectory: string = 'deps',
+    importMapFileName: string = 'main.importmap',
```
guybedford: I'd prefer if we could avoid this property. Generator operations are all import map in, import map out. So I think this one can follow the same process and not explicitly do the import map save but assume it is wrapped in a `inputMap` input and `getImportMap()` output. Better clarifying this "input / output" design of the generator would be nice, but it applies across all commands!

generator/src/generator.ts
```diff
@@ -1669,6 +1669,47 @@ export class Generator {
...
+    hostedBaseUrl: string = './'
+  ) {
+    if (!isNode) {
+      throw new JspmError('downloadAll is only available in nodejs');
+    }
+
+    const fs = await import('node:fs/promises');
+    const path = await import('node:path');
```
guybedford: We need to ensure these don't interfere with the browser build, I've been wrapping them in `eval()` to do that, there's likely a better way but worth noting.

generator/src/generator.ts
guybedford, line 1694: There should be a provider hook to get all the files of a package now I think - https://github.com/jspm/jspm/blob/main/generator/src/providers/index.ts#L45. It's not fully integrated yet, but that would be a nice way to capture this one.

guybedford: For iterating the packages, perhaps we could have a `generator.getPackageList` helper method to extract the package list from the generator of the form `{ registry, name, version, layer }[]`.

generator/src/generator.ts
```diff
@@ -1669,6 +1669,47 @@ export class Generator {
...
+
+    for (const url in this.traceMap.installer.resolver.traceEntries) {
+      const r = await fetch(url);
+      if (!r.ok) {
+        throw Error(`Failed to download file from url '${url}'`);
+      }
+      const text = await r.text();
+      const filename = path.resolve(outputDirectory, '.' + new URL(url).pathname);
```
guybedford, line 1701: It would be nice to support pattern replacements in outputDirectory e.g. `outputDirectory: 'jspm_packages/%registry/%layer/%name@%version'` something like that. This way we could replace the values in per package, allowing the directory to vary across packages.

guybedford (commented):
Left some initial suggestions, as I say let's discuss further on discord if possible.

generator/src/generator.ts
internettrans: I'm happy to remove `importMapFileName` and saving the import map to disk. What do you mean by "import map in, import map out?" I see that some methods call `this.importMap.clone()` rather than modifying `this.importMap` - is that related? Within `downloadAll`, I think the only change to the import map is to rebase the URL away from the registry urls to ones relative to `hostedBaseUrl`. Perhaps `downloadAll` should return that cloned, modified import map?

generator/src/generator.ts
internettrans: Is the browser build done by user code that bundles the published code? Running `chomp build` without the `eval()` results in a runtime `import()` that would be ignored if executed as-is within the browser. However, webpack/other bundlers would process if jspm generator is bundled by user code.

I'll change it to `eval` like you suggested, to be consistent. As an alternative, would moving the module names into variables avoid them being bundled in the major bundlers? Not sure how sophisticated they are nowadays.

```js
const modules = ['node:fs/promises', 'node:path'];
const [fs, path] = await import(modules);
```

generator/src/generator.ts
internettrans: I agree this feature would be convenient. It would require manipulation of the urls in the import map beyond what `rebase()` provides.

Are you open to an initial implementation that maintains the URL paths from the import map, with perhaps a follow-up to support rewriting the url paths?

internettrans:
I've updated this with feedback and believe this is mergeable. The failed tests are unrelated to this change.

I tried using `generator.getPackageList` but it requires passing in each pkgUrl (which is what `downloadAll` is looking for), and sometimes returns a falsey value. The `traceEntries` property within the resolver seems to have all urls within the import map, at least from the packages I've tested.

I verified manually that the files downloaded to the test fixture are loadable in Firefox, with the import map working. I also refined the implementation to support other providers besides the default jspm provider. The automated test is simple but verifies that some of the most popular npm packages can be used with the API.

This implementation is simple and doesn't implement the directory-rewriting that you suggested - happy to add that in a follow-up PR if there is interest in the feature. The OpenMRS ecosystem would use this API for self-hosting of an EMR, and the official single-spa examples would be updated to use it.

generator/src/generator.ts
guybedford: Rather than returning the cloned and modified import map, the current pattern in the generator is to actually mutate `generator.importMap`. Think of `generaator.importMap` as an exact mirror of the state of `importmap.json`. The JSPM CLI then wraps all generator calls with picking up the JSON file and saving it back when using a command.

Are there any reasons mutation can't work for your use case here?

generator/src/generator.ts
guybedford: The variables approach is also a good one, although it can become a game of cat and mouse with static analysis.

These days the better approach for Node.js builtins is actually to just use `process.getBuiltinModule`. I believe it might work for our Node.js compat these days. Can we use that here instead perhaps? Would be nice to replace the other usages of Node.js builtins with this approach too.

generator/src/generator.ts
guybedford: There were two feedback comments here:

1. Using a provider hook to get the list of files for a package to download
2. Extracting out `generator.getPackageList()` as an intermediate method to find the canonical package list used by the current import map.

I'm fine with skipping (2) above, but I do think we need to solve (1) for this PR, otherwise only downloading partial packages risks paths like asset loading and dynamic import being missed.

generator/src/generator.ts
guybedford: We can certainly add this feature as a follow-on.

guybedford:
Thanks @jolyndenning for the work on this. Let me know if you want to sync on the last remaining questions around a provider hook for file listing / package downloading here.

guybedford:
To clearly explain again what is missing here:

* Downloading an import map trace is a distinct operation from downloading a set of packages. JSPM should do full package downloads because modules can reference assets outside of the module graph trace. A stylesheet imported with `"type": "css"` can reference assets outside that stylesheet and won't reflect in the module trace.
* JSPM's goal for downloading is also to align with https://github.com/jspm/jspm/issues/2591 to allow local package downloads where those downloads include the whole packages, so an import map trace is not enough to satisfy a local package workflow where those package folders are considered constant and not traced subsets.

There are perhaps two separate use cases here:
* Import map trace and download, something more like you'd expect a download a webpage feature to work.
* Package list download, the JSPM local package management workflow.

A dedicated tool for tracing import maps and turning them into a local map and a folder with the downloaded contents sounds like a useful tool as well, but that is not suitable for JSPM.

guybedford:
Thinking some more about alternatives here - note that the install operation returns a `{ staticDeps, dynamicDeps }` pair, which can be used to determine the trace data.

Having APIs that better expose trace metadata to build this kind of tooling around JSPM could be an alternative approach to this PR - i.e. a `jspm.traceAll` instead of downloadAll just returning a manifest. Happy to discuss any variations here like that that might work further too.

internettrans:
Thanks for your reviews. This implementation approach doesn't look like the correct one, due to the module trace not containing all possible files that could be downloaded. The inner details of jspm generator are complex, such that I wasn't able to find another implementation approach. Let's discuss a provider hook for file listing - I'll message on discord.

Self hosting and open source are aggressed because cloud

PR #2706, opened 2026-03-02
fix: ensure 404s are never cached

guybedford:
I noticed sometimes 404s seem to get cached and not retry at the actual node fetch cache level.

This might help with that.


guybedford:
Superseded by https://github.com/jspm/jspm/pull/2713.

Issue #2702, opened 2026-02-28
Top-level wildcard exports can result in huge import maps


LeaVerou:
## Problem

   `combineSubpaths()` currently only combines subpath mappings in `scopes`, explicitly skipping top-level `imports`:
https://github.com/jspm/jspm/blob/3dc89cdd4cce50f8d606dab021d0e11d0fff2e74/import-map/src/map.ts#L310-L314 

   This causes wildcard exports like `"./modules/*": "./modules/*"` to produce hundreds of individual import map entries instead of a single trailing-slash mapping, when all subpaths are installed (via `subpaths: true`).

Import maps natively support trailing-slash (prefix) mappings, which are the natural representation of wildcard exports. Expanding them into individual entries produces import maps that are orders of magnitude larger than necessary. As a real-world example, underscore 1.13.7 has 5 wildcard export patterns (./underscore*, ./modules/*, ./amd/*, ./cjs/*) which expand to ~200 entries instead of 5 trailing-slash mappings.

This doesn't surface in typical JSPM usage because installed subpaths get pruned to only what's actually imported. But consumers of the generator API that install all subpaths (e.g. for offline/bundler-free workflows) get massively bloated maps.

## Reduced test case

First, 

```bash
npm i underscore
npm i @jspm/generator -D
```

Then, in `index.js`:

   ```js
import { Generator } from "@jspm/generator";
import { writeFileSync, existsSync } from "node:fs";

// Install underscore with all subpaths (as a consumer wanting a complete import map)
const generator = new Generator({
	defaultProvider: "nodemodules",
	env: ["production", "browser", "module"],
	// combineSubpaths defaults to true
});

if (!existsSync("./node_modules/underscore/modules/pipe.js")) {
	// Avoid bizarre error message about missing pipe.js that does not seem imported from anywhere.
	// Unclear whether this is a JSPM bug or an Underscore bug, but it's not relevant here
	// (and fixing this bug would also make it disappear)
	writeFileSync("./node_modules/underscore/modules/pipe.js", "", "utf-8");
}

await generator.install({
	alias: "underscore",
	target: "./node_modules/underscore",
	subpaths: true,
});

const map = generator.getMap();
const entryCount = Object.keys(map.imports).length;
console.log(`${entryCount} entries in imports`);
console.log(JSON.stringify(map, null, "\t"));
```

Then run `node index.js`

Actual output: ~500 individual entries, one per file in `modules/`, `amd/`, `cjs/`, and `underscore*`.
[...]

PR #2708, opened 2026-03-02
feat: Convert compatible wildcard exports to prefix mappings losslessly

LeaVerou:
Adds a new `expandWildcards` option: 
- `expandWildcards: false` is the current behavior, 
- `expandWildcards: true` losslessly converts _compatible_ wildcard exports to prefixes as follows: 
	1. Trailing-wildcard exports (e.g. `pkg/*` → `./path/*`) are _always_ converted to prefixes
	2. Wildcard exports with the same prefix (e.g. `pkg/*.js` → `./path/*.js`) are _only_ converted to prefix mappings iff there are no other files in that path
	3. Other wildcard exports are not affected (they are expanded per the previous behavior)

This process is lossless because it preserves the intent of the original export and does not expose any additional files. Therefore, the current default is `false`, as it could be considered a bugfix, but it's easy to change to `true` for compat (or even start off with `true` and convert to `false` in the next major version. Just let me know and I can change it!

For the `underscore` testcase in #2702, this PR would produce:

With `combineSubpaths: false, expandWildcards: false`:

```js
{
        "imports": {
                "underscore": "./node_modules/underscore/modules/index-all.js",
                "underscore/package.json": "./node_modules/underscore/package.json",
                "underscore/underscore": "./node_modules/underscore/underscore",
                "underscore/modules/": "./node_modules/underscore/modules/",
                "underscore/amd/": "./node_modules/underscore/amd/",
                "underscore/cjs/": "./node_modules/underscore/cjs/"
        }
}
```

With `combineSubpaths: "both", expandWildcards: false`:

```json
{
        "imports": {
                "underscore": "./node_modules/underscore/modules/index-all.js",
                "underscore/": "./node_modules/underscore/"
        }
}
```

## Implementation

- `expandExportsTarget()` detects collapsible wildcard patterns via a new `trailingSlashSubpaths` parameter
- New `ImportMap.prototype.condenseImports(prefixes)` replaces matching entries sharing a consistent URL base with a single trailing-slash entry
- `Generator.prototype.simplify()` encapsulates both simplifications and applies wildcard condensing before `combineSubpaths()`. I can get rid of this if you'd prefer and just change the three call sites.

### What about sharing code between `condenseImports()` and `combineSubpaths()`?

I wondered about sharing part of the implementation between the two since they seem to be doing similar things (`combineSubpaths` greedily, `condenseImports` for predefined prefixes) but it seems the two methods solve fundamentally different problems and thus there's not really a meaningful shared abstraction — extracting one would just move code around without reducing complexity.                                                                                            
  - `combineSubpaths()` does greedy discovery — it scans all entries and tries to find common prefixes by progressively shortening paths. It doesn't know which prefixes are valid; it just merges whatever it can. That's why it's lossy.                                                                                                                                                                                   
  - `condenseImports(prefixes)` does targeted condensing — it receives specific prefixes already identified as wildcard patterns during export expansion. It only touches those.                                  
                                                                                                                                                                                                                
A shared utility would need to go one of two ways:                                                                                                                                                            
1. Accept a set of prefixes → but then `combineSubpaths()` would need to discover prefixes first, which is most of its logic                                                                                      
2. Be greedy → but then `condenseImports()` loses its losslessness guarantee                                                                                                                                      

The discovery logic in `combineSubpaths` (progressively shortening paths to find a common base) is structurally different from the verification logic in `condenseImports` (check entries under a known prefix).

### What about a single setting for both?

That said, it _could_ make sense to expose a unified API and have a single setting (name & values TBD), with the following "levels":
1. Never combine anything, expand all wildcards
2. Only use prefixes for trailing-wildcard exports, expand everything else
3. Only use prefixes for trailing-wildcard exports, and compatible suffixes (the current `expandWildcards: false` that this PR introduces)
4. All of the above, plus post-hoc combining of everything we can combine (the current `combineSubpaths: true`)

## Things tested

- [x] Basic wildcard condensing (`./modules/*` → trailing-slash entry)
- [x] Works with both combineSubpaths: true and false
- [x] Shadowed exports preserved alongside trailing-slash entry
- [x] Suffix wildcards (`*.js` → `*.js`) condense correctly
- [x] Mixed file types: suffix wildcards with non-matching files in target dir are NOT condensed (no file leakage)
- [x] Manual test with underscore (`subpaths: true`)
- [ ] Other providers (so far only tested with `nodemodules`)

🤖 Co-authored with [Claude Code](https://claude.com/claude-code). 
👩🏽‍💻 Reviewed, iterated on, and tested by this human (except for the tests, I only skimmed those)

### DIFF
--- a/generator/src/common/package.ts
+++ b/generator/src/common/package.ts
@@ -54,7 +54,8 @@ export function expandExportsResolutions(
   exports: ExportsTarget | Record<string, ExportsTarget>,
   env: string[],
   files?: Set<string> | undefined,
-  exportsResolutions: Map<string, string> = new Map()
+  exportsResolutions: Map<string, string> = new Map(),
+  trailingSlashSubpaths?: Set<string>
 ) {
   if (typeof exports !== 'object' || exports === null || !allDotKeys(exports)) {
     let targetList = new Set<string>();
@@ -75,7 +76,8 @@ export function expandExportsResolutions(
           subpath,
           target,
           files,
-          exportsResolutions
+          exportsResolutions,
+          trailingSlashSubpaths
         );
       }
     }
@@ -203,7 +205,8 @@ function expandExportsTarget(
   subpath: string,
   target: string,
   files: Set<string> | undefined,
-  entriesMap: Map<string, string>
+  entriesMap: Map<string, string>,
+  trailingSlashSubpaths?: Set<string>
 ) {
   if (!target.startsWith('./') || !(subpath.startsWith('./') || subpath === '.')) return;
   if (target.indexOf('*') === -1 || subpath.indexOf('*') === -1) {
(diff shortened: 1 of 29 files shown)


guybedford:
I really like the concept here in using the original wildcards to drive the optimization.

I've added two refactorings to this PR:

1. Adding support for scopes in condense by supporting a imports and scopes lists instead of a singular list
2. Caching the file list in the resolver and then lazily populating the prefix data when needed

I think it can make sense to apply this to scopes if you consider dynamic import cases like `import(`#foo/${feature}`)` inside of a utility library say, where feature is a user-provided string. While less important, it seems like it can apply just as well.

guybedford (approved):

PR #2713, opened 2026-03-30
feat: @jspm/fetch

guybedford:
This implements a new `@jspm/fetch` package which offers significant performance improvements over the `node-fetch-cache` package, including resolving https://github.com/jspm/jspm/issues/2710.

Uses a SQLite caching layer on the filesystem with less lock contention than cacache.

All failures are due to `@jspm/fetch` not yet being published on npm.


guybedford (commented):
## Review: @jspm/fetch

### Architecture
Extracts the fetch/cache layer from `@jspm/generator` into a new `@jspm/fetch` workspace package. Replaces `node-fetch-cache` (file-system cacache-based) with a SQLite-backed two-tier cache (memory + SQLite with Brotli compression). Clean separation of concerns.

### Issues

**1. `better-sqlite3` is a native addon — this is a significant dependency change.**
It requires a C++ compiler or prebuilt binaries at install time. This will break environments where native compilation isn't available (some CI, restricted containers, Stackblitz, etc.). The old `node-fetch-cache` was pure JS. This should be a deliberate, documented tradeoff.

**2. `CachedResponseImpl.text()` and `.json()` are synchronous, but `WrappedResponse` declares them as returning `Promise<string> | string`.**
The generator code throughout does `await res.text()` and `await res.json()`. Returning sync values works with `await` (it just unwraps), but the `dirResponse` in `protocols.ts:613` returns `_readdir(path)` from `.json()` — that's a `Promise<Dirent[]>`, not `string[]`. The caller needs to `await` it. This works because the generator already awaits, but the type signature (`json(): any`) hides this mismatch.

**3. `handleFileUrl` (protocols.ts) has inverted logic for directory detection.**
```ts
if (url.endsWith('/')) {
  try {
    readFileSync(path);  // succeeds = it's a file, not a dir
    return new CachedResponseImpl(url, 404, 'Directory does not exist', ...);
```
If `readFileSync` succeeds on a path ending with `/`, that means a file was read at that path — returning 404 "Directory does not exist" is confusing. The intent is to detect whether the path is a directory, but the error message and logic are inverted from the old code which had the same pattern. It works because paths ending in `/` almost always hit `EISDIR`, but it's fragile.

**4. `setRetryCount` and `setFetchPoolSize` are now no-ops.**
These were public API on `@jspm/generator`. Making them no-ops is a breaking change for anyone who configured retry/pool behavior. The comment says "for compat" but they silently do nothing.

**5. No HTTP cache header handling.**
The old `node-fetch-cache` respected HTTP cache semantics (Cache-Control, ETag, etc.). The new cache stores responses indefinitely (or until manually cleared). The `immutable` flag exists but `FetchOptions.cache` modes (`no-cache`, `force-cache`) aren't fully implemented — only `no-store` is checked. For a package manager, stale cache entries could cause real issues.

**6. Pool scheduling changed from LIFO to FIFO.**
Old `Pool.pop()` used `Array.pop()` (LIFO), new `Pool.release()` uses `Array.shift()` (FIFO). FIFO is arguably better (fairer), but it's a silent behavioral change.

**7. `--test-force-exit` added to all test runners in chompfile.**
This could mask test cleanup issues (dangling handles, unclosed connections). Is this needed because of SQLite connections not closing?

### What looks good
- SQLite WAL mode + `synchronous=NORMAL` is the right choice for a cache
- Brotli at quality 1 (fast) is sensible for cache compression
- Two-tier memory+disk cache design is clean
- In-flight request deduplication is a nice addition
- Virtual source handling was cleanly migrated

guybedford:
Replaced with https://github.com/jspm/jspm/pull/2729.

PR #2728, opened 2026-04-26
fix: parallelize per-module dependency visits in tracemap

### LINKED ISSUE #2727: Generator does serial fetches during dependency resolution — N × RTT install time
When `Generator.install()` is called from a network location with non-trivial RTT to `ga.jspm.io`, package resolution issues fetches strictly one-at-a-time. With N modules in the install closure, install latency scales linearly with module count rather than dependency-tree depth.

For example, installing `core-js@3.49.0` over a connection with ~150ms RTT to the CDN takes ~7.6s wall-clock for 39 modules.

### Reproduction

```js
import { Generator } from '@jspm/generator';
const g = new Generator({
  defaultProvider: 'jspm.io',
  env: ['browser', 'production', 'module']
});
console.time('install');
await g.install('core-js@3.49.0');
console.timeEnd('install');
```

### What the HAR shows

- 39 modules fetched, **max concurrent in flight = 1** for the entire run
- HTTP/2 over a single connection (multiplexing available, not used)
- ~2ms gap between each finish and the next request — too fast to be browser-side delay
- Total wall ≈ 7.6s

Every request's `_initiator.stack` traces back to `@jspm/generator`'s own bundle, with a recurring frame chain:

```
visit → resolve → finalizeResolve → exists → analyze → _analyzeAsync → getAnalysis → fetch
```

### Expected behavior

Per-module analysis should fan out across a module's imports. The classic shape is `await Promise.all(deps.map(d => visit(d)))` rather than `for (const d of deps) await visit(d)`. With proper parallelism, install time should be roughly `depth × RTT` instead of `count × RTT` — typically a 5–10× speedup for a wide tree like core-js.

### Proof that browser-native ESM is not the bottleneck

The same package's modules, loaded inside an iframe via native `<script type=\"module\">` after install, show **max concurrent = 33** and complete in ~100ms total. The parallelism issue is specifically in the Generator's resolution loop, not in the browser, the network protocol, or the CDN.

### Environment

- `@jspm/generator` 2.13.0
- Chromium DevTools HAR
- HTTP/2, single TCP connection, no protocol-level serialization

Happy to share the HAR file if useful.

guybedford:
Closes #2727.

`Generator.install()` was issuing CDN fetches strictly one-at-a-time during dependency resolution, because `TraceMap.visit` walked sibling deps via `yield* this._visit(dep, ...)`, making install latency scale with module count × RTT instead of dependency-tree depth × RTT. An earlier revision of this PR fixed the parallelism by converting `visit` to a plain `async` function, but that regressed the warm-cache path: every visit incurred an `await` microtask even when nothing needed I/O.

This version restores the fast path while keeping the parallel cold-path fan-out.

- `visit()` now returns `string | null | undefined | Promise<...>` instead of always-Promise. Cache-warm sub-trees walk synchronously with zero microtask overhead.
- The three potentially-async transition points (resolve, analyze, async visitor) each split off into a small `async` helper that re-enters the sync continuation. If all three caches hit and there is no visitor, the entire subtree completes without ever creating a Promise.
- The dep fan-out collects child return values, builds `Promise.all` only over the ones that were Promises, and returns the parent's resolved URL synchronously when none were.
- Mirrors the same `T | Promise<T>` pattern already used by `Resolver.analyze`.

Before — every level always awaits, sibling deps serialised by the generator's `yield*`:

```ts
async visit(...) {
  resolved = await this.resolve(...);          // microtask even on cache hit (pre-async-rewrite)
  entry = await this.resolver.analyze(...);    // ditto
  await Promise.all(allDeps.map(dep => this.visit(dep, ...)));
}
```

After — sync when warm, parallel when cold, no wasted microtasks for cached siblings:

```ts
visit(...): string | null | undefined | Promise<...> {
  // ... sync cache lookups for resolve + analyze ...
  if (resolved === undefined) return this._visitFromResolve(...); // async branch
  return this._afterResolve(...);                                  // stays sync
}

private _fanout(...) {
  let pending: Promise<unknown>[] | undefined;
  for (const dep of allDeps) {
    const r = this.visit(dep, opts, resolved, seen);
    if (r && typeof (r as Promise<unknown>).then === 'function') {
      (pending ??= []).push(r as Promise<unknown>);
    }
  }
  if (!pending) return resolved;
  return Promise.all(pending).then(() => resolved);
}
```

Cross-branch dedup is still preserved: `seen` is mutated synchronously before any await, so two parallel visits to the same `${specifier}##${parentUrl}` key still race-resolve to a single visit.

### DIFF
--- a/generator/src/trace/tracemap.ts
+++ b/generator/src/trace/tracemap.ts
@@ -177,6 +177,11 @@ export default class TraceMap {
   /**
    * Resolves, analyses and recursively visits the given module specifier and all of its dependencies.
    *
+   * Per-module dependencies are walked in parallel via `Promise.all`, so a wide
+   * dependency tree's wall-clock time scales with depth × RTT rather than the
+   * total module count × RTT. The `seen` set is mutated synchronously before any
+   * await, which preserves cross-branch deduplication under the parallel fan-out.
+   *
    * @param {string} specifier Module specifier to visit.
    * @param {VisitOpts} opts Visitor configuration.
    * @param {} parentUrl URL of the parent context for the specifier.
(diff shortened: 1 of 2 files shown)


guybedford:
Turns out this slows down the perf case this codepath fixed of course.

So we likely need to do a hybrid perf + deopt approach.

PR #2739, opened 2026-05-31
fix broken link to CLI docs

njbair:
Hi all, thanks so much for JSPM and your work on advancing web standards.

The link to the CLI docs in README is broken. This updates with the correct link.

I searched through the code to make sure this was the only instance.

Thanks again.


guybedford:
Sure, thanks for posting the PR.

